PR #10, opened 2016-12-20
Add tool-friendly intermediate ASTs

### LINKED ISSUE #9: Authors in pubspec should be standard Dart owners
See other packages – likely should copy pkg/angular2

matanlurey:
Also closes https://github.com/dart-lang/angular_ast/issues/9

### DIFF
--- a/lib/src/ast/sugar/banana.dart
+++ b/lib/src/ast/sugar/banana.dart
@@ -1,0 +1,96 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'package:angular_ast/src/ast.dart';
+import 'package:angular_ast/src/token.dart';
+import 'package:source_span/source_span.dart';
+import 'package:quiver/core.dart';
+
+/// Represents the `[(property)]="field"` syntax.
+///
+/// This AST may only exist in the parses that do not de-sugar directives (i.e.
+/// useful for tooling, but not useful for compilers).
+///
+/// Clients should not extend, implement, or mix-in this class.
+abstract class BananaAst implements TemplateAst {
+  /// Create a new synthetic [BananaAst] with a string [field].
+  factory BananaAst(
+    String name, [
+    String field,
+  ]) = _SyntheticBananaAst;
+
+  /// Create a new synthetic [BananaAst] that originated from node [origin].
+  factory BananaAst.from(
+    TemplateAst origin,
+    String name, [
+    String field,
+  ]) = _SyntheticBananaAst.from;
+
+  /// Create a new [BananaAst] parsed from tokens from [sourceFile].
+  factory BananaAst.parsed(
+    SourceFile sourceFile,
+    NgToken nameToken,
+    NgToken fieldToken,
+    NgToken endFieldToken,
+  ) = _ParsedBananaAst;
+
+  @override
+  bool operator ==(Object o) {
+    if (o is BananaAst) {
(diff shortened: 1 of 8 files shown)


lib/src/ast/sugar/banana.dart
```diff
@@ -0,0 +1,96 @@
...
+
+  @override
+  String toString() {
+    return '$BananaAst {$name="$field"}';
+  }
+}
+
+class _ParsedBananaAst extends TemplateAst with BananaAst {
+  final NgToken _nameToken;
+  final NgToken _fieldToken;
```
MichaelRFairhurst, line 62: Is this an NgToken or a dart.AstNode? Maybe a NgToken right now and dart parsing to come later?

lib/src/ast/sugar/star.dart
```diff
@@ -0,0 +1,101 @@
...
+    SourceFile sourceFile,
+    NgToken beginToken,
+    this._nameToken,
+    NgToken endToken, [
+    this.expression,
+  ])
+      : super.parsed(beginToken, endToken, sourceFile);
+
+  @override
+  final ExpressionAst expression;
```
MichaelRFairhurst, line 77: I think this needs to be a more complicated data type.

Like, `*ngFor="let item of items; let i=index; trackBy: myStuff"` should result in a list of lets and input bindings. For when it is an expression like in ngIf, it can create an input binding where the name is ngIf and the NgToken is the token after the star.

lib/src/parser/recursive.dart
```diff
@@ -210,8 +218,14 @@ class RecursiveAstParser {
       references: references,
     );
     if (deSugarTemplateAst != null) {
-      return new EmbeddedTemplateAst.from(
+      TemplateAst origin = deSugarTemplateAst;
+      origin = new StarAst.from(
```
MichaelRFairhurst, line 222: This doesn't seem to support `template="ngFor ...` yet, is that intentional, or am I wrong?

MichaelRFairhurst (approved):
Great! Definitely looks like it will be useful to have this info. Some comments on collecting more, but that's probably just because this isn't supposed to be production ready yet. But maybe useful reminders!

lib/src/ast/sugar/star.dart
matanlurey: Hmm. And this is just for `*` right, not `<template>` (desugared) right?

I imagine then we could have a `StarExpressionAst` which is more complex (and itself can be desugared into multiple expressions). WDUT?

lib/src/parser/recursive.dart
matanlurey: What is `template="ngFor ..."` I actually don't know what that is...

lib/src/ast/sugar/banana.dart
matanlurey: It's a raw Token.

The current plan is to port the current expression parser and use it to parse lazily.

i.e. You'll do

```dart
// You now have an `ExpressionAst` object, which has been parsed from the string.
var expressionAst = bananaAst.expression;

// Do fun things
expressionAst.accept(...)
```

lib/src/ast/sugar/star.dart
MichaelRFairhurst: Sounds gtm

lib/src/parser/recursive.dart
MichaelRFairhurst: Simple enough: *x="y" and template="x y" are equivalent

matanlurey: Oh boy. I'd love to kill the latter.

PR #11, opened 2016-12-22
Add expression support

matanlurey:
Special thanks to @mk13 and @MichaelRFairhurst for inspiration to do it this way!

/cc @jonahwilliams

### DIFF
--- a/lib/src/ast/expression.dart
+++ b/lib/src/ast/expression.dart
@@ -2,21 +2,41 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// TODO: This is a placeholder.
-//
-// Probably want to either:
-// - Port the original one
-// - Write a new one
-// - Use the Dart analyzer and special case pipes
-// - Use the Dart analyzer and transform pipes
-class ExpressionAst {
-  final String expression;
+import 'package:analyzer/analyzer.dart';
+import 'package:angular_ast/src/expression/parser.dart';
 
+/// Wraps a parsed Dart [Expression] as an Angular [ExpressionAst].
+class ExpressionAst {
+  /// Dart expression.
+  final Expression expression;
+
+  /// Create a new expression AST wrapping a Dart expression.
   const ExpressionAst(this.expression);
 
+  /// Create a new expression AST by parsing [expression].
+  factory ExpressionAst.parse(
+    String expression, {
+    bool deSugarPipes: true,
+    String sourceUrl: '/test.dart',
+  }) {
+    // Fall back in case this comes from a test w/o a real source URL.
+    if (!sourceUrl.endsWith('.dart')) {
+      sourceUrl = '/test.dart';
+    }
+    return new ExpressionAst(parseExpression(
+      expression,
+      deSugarPipes: deSugarPipes,
+      sourceUrl: sourceUrl,
+    ));
+  }
+
   @override
-  bool operator ==(Object o) =>
-      o is ExpressionAst && o.expression == expression;
+  bool operator ==(Object o) {
+    if (o is ExpressionAst) {
+      return o.expression.toSource() == expression.toSource();
+    }
+    return false;
+  }
 
   @override
   int get hashCode => expression.hashCode;
--- a/lib/src/expression/parser.dart
+++ b/lib/src/expression/parser.dart
@@ -1,0 +1,128 @@
+import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/standard_ast_factory.dart';
+import 'package:analyzer/error/listener.dart';
+import 'package:analyzer/file_system/memory_file_system.dart';
+import 'package:analyzer/src/dart/ast/token.dart';
+import 'package:analyzer/src/dart/scanner/reader.dart';
+import 'package:analyzer/src/dart/scanner/scanner.dart';
+import 'package:analyzer/src/generated/parser.dart';
+import 'package:analyzer/src/generated/source.dart';
+
+final _resourceProvider = new MemoryResourceProvider();
+
+class _ThrowingListener implements AnalysisErrorListener {
+  const _ThrowingListener();
+
+  @override
+  void onError(AnalysisError error) {
+    throw new FormatException(
+      error.toString(),
+      error.source.contents.data,
+      error.offset,
+    );
+  }
+}
+
+/// Parses a template [expression].
+Expression parseExpression(
+  String expression, {
+  bool deSugarPipes: true,
+  String sourceUrl: '/~test.dart',
+}) {
+  final source = _resourceProvider
+      .newFile(
+        sourceUrl,
+        expression,
+      )
+      .createSource();
+  final reader = new CharSequenceReader(expression);
+  final listener = const _ThrowingListener();
+  final scanner = new Scanner(source, reader, listener);
+  final parser = new _NgExpressionParser(
+    source,
+    listener,
+    deSugarPipes: deSugarPipes,
+  );
+  return parser.parseExpression(scanner.tokenize());
+}
+
+/// Extends the Dart language to understand the current Angular 'pipe' syntax.
+///
+/// Based on https://github.com/dart-lang/angular_analyzer_plugin/pull/160
+class _NgExpressionParser extends Parser {
+  final bool _deSugarPipes;
+
+  _NgExpressionParser(
+    Source source,
+    AnalysisErrorListener errorListener, {
+    bool deSugarPipes,
+  })
+      : _deSugarPipes = deSugarPipes,
+        super(source, errorListener);
+
+  @override
+  Expression parseBitwiseOrExpression() {
+    if (!_deSugarPipes) {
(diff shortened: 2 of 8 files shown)


lib/src/ast/expression.dart
```diff
@@ -2,21 +2,41 @@
...
-// - Use the Dart analyzer and special case pipes
-// - Use the Dart analyzer and transform pipes
+import 'package:analyzer/analyzer.dart';
+import 'package:angular_ast/src/expression/parser.dart';
+
+/// Wraps a parsed Dart [Expression] as an Angular [ExpressionAst].
 class ExpressionAst {
-  final String expression;
+  /// Dart expression.
+  final Expression expression;
```
MichaelRFairhurst, line 11: just from a readability standpoint, importing this as dart and using `dart.Expression` may be a hair clearer

lib/src/ast/expression.dart
```diff
@@ -2,21 +2,41 @@
...
   const ExpressionAst(this.expression);
 
+  /// Create a new expression AST by parsing [expression].
+  factory ExpressionAst.parse(
+    String expression, {
+    bool deSugarPipes: true,
+    String sourceUrl: '/test.dart',
+  }) {
+    // Fall back in case this comes from a test w/o a real source URL.
+    if (!sourceUrl.endsWith('.dart')) {
```
MichaelRFairhurst, line 23: Hmm, this seems like mixing testing concerns with production code

lib/src/expression/parser.dart
```diff
@@ -0,0 +1,128 @@
...
+  String sourceUrl: '/~test.dart',
+}) {
+  final source = _resourceProvider
+      .newFile(
+        sourceUrl,
+        expression,
+      )
+      .createSource();
+  final reader = new CharSequenceReader(expression);
+  final listener = const _ThrowingListener();
```
MichaelRFairhurst, line 41: This won't work for us, we want to still validate code that was successfully parsed even if some of it failed

lib/src/expression/parser.dart
```diff
@@ -0,0 +1,128 @@
...
+      : _deSugarPipes = deSugarPipes,
+        super(source, errorListener);
+
+  @override
+  Expression parseBitwiseOrExpression() {
+    if (!_deSugarPipes) {
+      return super.parseBitwiseOrExpression();
+    }
+    Expression expression;
+    if (currentToken.keyword == Keyword.SUPER &&
```
MichaelRFairhurst, line 71: what's SUPER doing here?

lib/src/expression/parser.dart
```diff
@@ -0,0 +1,128 @@
...
+  /// Given an expression of `{expression} | {pipe}`, returns an expression.
+  ///
+  /// For example, will return:
+  /// ```
+  /// $$ng.pipe({pipe}, {expression}, [{args}])
+  /// ```
+  ///
+  /// With the expectation the compiler will optimize further.
+  Expression parseAndTransformPipeExpression(BinaryExpression expression) {
+    final ngInternalNamespace = astFactory.simpleIdentifier(
```
MichaelRFairhurst, line 97: I think this should be PrefixedIdentifier

lib/src/expression/parser.dart
```diff
@@ -0,0 +1,128 @@
...
+    }
+    Expression expression;
+    if (currentToken.keyword == Keyword.SUPER &&
+        currentToken.next.type == TokenType.BAR) {
+      expression = new SuperExpression(getAndAdvance());
+    } else {
+      expression = parseBitwiseXorExpression();
+    }
+    while (currentToken.type == TokenType.BAR) {
+      expression = new BinaryExpression(
```
MichaelRFairhurst, line 78: Rather than creating a BinaryExpression and breaking it apart later, I would just getAndAdvance(), parseBitwiseXorExpression() and pass "expression, pipeExpression" into the next function

MichaelRFairhurst: Also, the resulting AST will refer to "pipe" as a var within scope. Is it a var within the dart scope? Ie, assuming "$ng.pipe" were a real function, would `$ng.pipe(expr, myPipe)` actually be a valid expression, or would it require like `$ng.pipe(expr, 'myPipe`)`?

lib/src/expression/parser.dart
```diff
@@ -0,0 +1,128 @@
...
+    );
+  }
+}
+
+/// Parses a template [expression].
+Expression parseExpression(
+  String expression, {
+  bool deSugarPipes: true,
+  String sourceUrl: '/~test.dart',
+}) {
```
MichaelRFairhurst, line 33: This needs offset data

test/expression/parser_test.dart
```diff
@@ -0,0 +1,58 @@
...
+      'foo || bar',
+    ].forEach((expression) {
+      expect(parseExpression(expression).toSource(), expression);
+    });
+  });
+
+  test('should parse a simple pipe', () {
+    expect(
+      parseExpression('foo | bar').toSource(),
+      r'$$ng.pipe.bar(foo)',
```
MichaelRFairhurst, line 25: Ah, I see what you're doing here.

I think in this case you should make sure the parser only accepts identifiers as a pipename, rather than...I think the current implementation accepts BinaryXorExpression (which could mean `$ng.pipe.(any + stuff + in + here)(foo)`)

MichaelRFairhurst (commented):
Really cool solution right? This looks like an awesome step, though could use more work in terms of what we will need out of it (offsets, graceful failure). And it looks like its a bit too accepting right now which is the good side to err on but could lead to some weird stuff.

lib/src/expression/parser.dart
```diff
@@ -0,0 +1,128 @@
...
+  /// ```
+  /// $$ng.pipe({pipe}, {expression}, [{args}])
+  /// ```
+  ///
+  /// With the expectation the compiler will optimize further.
+  Expression parseAndTransformPipeExpression(BinaryExpression expression) {
+    final ngInternalNamespace = astFactory.simpleIdentifier(
+      new StringToken(TokenType.IDENTIFIER, r'$$ng.pipe', 0),
+    );
+    // TODO: Support pipe arguments.
```
MichaelRFairhurst, line 100: do you mean TODO support named arguments?

matanlurey: Acknowledged. Will remove this comment in follow-up.

lib/src/expression/parser.dart
matanlurey: Acknowledged. Will remove this comment in follow-up:

https://github.com/dart-lang/angular_ast/issues/14

test/expression/parser_test.dart
matanlurey: Acknowledged. Will remove this comment in follow-up:

https://github.com/dart-lang/angular_ast/issues/14

lib/src/expression/parser.dart
matanlurey: I think the idea here is that the compiler will not invoke `$$ng` prefixed identifiers, and instead use it as a signal that custom code needs to be created - sort of an annotation if you will. So no direct invocation.

lib/src/expression/parser.dart
matanlurey: Copied and pasted from the base `Parser`, I realize it doesn't make sense.

There will likely need to be a secondary check at resolution time to look for invalid keywords, missing identifiers, etc, and it can't be here.

lib/src/expression/parser.dart
matanlurey: Acknowledged. Will fix this in follow-up:

https://github.com/dart-lang/angular_ast/issues/14

lib/src/expression/parser.dart
matanlurey: Acknowledged. Will fix this in follow-up:

https://github.com/dart-lang/angular_ast/issues/14

lib/src/ast/expression.dart
matanlurey: Yeah, it's just a hack - need to make this better. Filed #14.

lib/src/ast/expression.dart
matanlurey: Acknowledged. Will fix this in follow-up:

https://github.com/dart-lang/angular_ast/issues/14

lib/src/expression/parser.dart
MichaelRFairhurst: Hmm, I think the only thing about that to me is that the resulting AST has a semantic meaning, just one that existing tools won't like.

I think it'd be best to transform the semantic meaning into a valid one (ie, actually expose $$ng.pipe to users and everything), or create an extension to the AST so that existing tools know that they need to be extended with `visitPipe` etc before they can work.

lib/src/expression/parser.dart
MichaelRFairhurst: named arguments aren't supported in the code yet, right? Don't think I saw them

matanlurey: Yes, they are not supported.

MichaelRFairhurst (approved):
Approved since all secondary issues are filed

lib/src/expression/parser.dart
matanlurey: Hmm, I could create a `PipeExpression` extension pretty simply.

Let's look @ https://github.com/dart-lang/angular_ast/issues/16

PR #12, opened 2016-12-23
Add TemplateAstVisitor, parse *ngFor micro expressions

matanlurey:
Branched from https://github.com/dart-lang/angular_ast/pull/11, which I'm waiting on review for.

### DIFF
--- a/lib/src/ast/attribute.dart
+++ b/lib/src/ast/attribute.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'package:angular_ast/src/ast.dart';
 import 'package:angular_ast/src/token.dart';
+import 'package:angular_ast/src/visitor.dart';
 import 'package:source_span/source_span.dart';
 import 'package:quiver/core.dart';
 
@@ -32,8 +33,17 @@ abstract class AttributeAst implements TemplateAst {
   ]) = _ParsedAttributeAst;
 
   @override
-  bool operator ==(Object o) =>
-      o is AttributeAst && name == o.name && value == o.value;
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
+    return visitor.visitAttribute(this, context);
+  }
+
+  @override
+  bool operator ==(Object o) {
+    if (o is AttributeAst) {
+      return name == o.name && value == o.value;
+    }
+    return false;
+  }
 
   @override
   int get hashCode => hash2(name, value);
--- a/lib/src/ast/comment.dart
+++ b/lib/src/ast/comment.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'package:angular_ast/src/ast.dart';
 import 'package:angular_ast/src/token.dart';
+import 'package:angular_ast/src/visitor.dart';
 import 'package:source_span/source_span.dart';
 
 /// Represents a comment block of static text.
@@ -25,6 +26,11 @@ abstract class CommentAst implements StandaloneTemplateAst {
     NgToken valueToken,
     NgToken endCommentToken,
   ) = _ParsedCommentAst;
+
+  @override
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
+    return visitor.visitComment(this, context);
+  }
 
   @override
   bool operator ==(Object o) => o is CommentAst && value == o.value;
--- a/lib/src/ast/expression.dart
+++ b/lib/src/ast/expression.dart
@@ -2,21 +2,41 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// TODO: This is a placeholder.
-//
-// Probably want to either:
-// - Port the original one
-// - Write a new one
-// - Use the Dart analyzer and special case pipes
-// - Use the Dart analyzer and transform pipes
-class ExpressionAst {
-  final String expression;
+import 'package:analyzer/analyzer.dart';
+import 'package:angular_ast/src/expression/parser.dart';
 
+/// Wraps a parsed Dart [Expression] as an Angular [ExpressionAst].
+class ExpressionAst {
+  /// Dart expression.
+  final Expression expression;
+
+  /// Create a new expression AST wrapping a Dart expression.
   const ExpressionAst(this.expression);
 
+  /// Create a new expression AST by parsing [expression].
+  factory ExpressionAst.parse(
+    String expression, {
+    bool deSugarPipes: true,
+    String sourceUrl: '/test.dart',
+  }) {
+    // Fall back in case this comes from a test w/o a real source URL.
+    if (!sourceUrl.endsWith('.dart')) {
+      sourceUrl = '/test.dart';
+    }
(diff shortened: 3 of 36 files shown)


lib/src/ast/attribute.dart
```diff
@@ -32,8 +33,17 @@ abstract class AttributeAst implements TemplateAst {
   ]) = _ParsedAttributeAst;
 
   @override
-  bool operator ==(Object o) =>
-      o is AttributeAst && name == o.name && value == o.value;
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
```
MichaelRFairhurst, line 36: is C supposed to still be here?

lib/src/ast/comment.dart
```diff
@@ -27,6 +28,11 @@ abstract class CommentAst implements StandaloneTemplateAst {
   ) = _ParsedCommentAst;
 
   @override
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
```
MichaelRFairhurst, line 31: Same comment on `C`

lib/src/ast/expression.dart
```diff
@@ -2,21 +2,41 @@
...
+
   @override
-  bool operator ==(Object o) =>
-      o is ExpressionAst && o.expression == expression;
+  bool operator ==(Object o) {
+    if (o is ExpressionAst) {
+      return o.expression.toSource() == expression.toSource();
+    }
+    return false;
+  }
```
MichaelRFairhurst, line 39: Could be useful to add to the visitor as well (for instance, we'll probably put a scope of local variables on each of these, and then run a visitor that just resolves every Expression. Might not want to have to implement visitMustache, visitInput, etc etc just to get at all the Exrpessions).

lib/src/ast/sugar/banana.dart
```diff
@@ -35,6 +36,11 @@ abstract class BananaAst implements TemplateAst {
   ) = _ParsedBananaAst;
 
   @override
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
+    throw new UnimplementedError();
```
MichaelRFairhurst, line 40: Works for now but I'm hoping this changes! :)

lib/src/ast/sugar/star.dart
```diff
@@ -36,6 +37,11 @@ abstract class StarAst implements TemplateAst {
   ]) = _ParsedStarAst;
 
   @override
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
+    throw new UnimplementedError();
```
MichaelRFairhurst, line 41: same comment as Banana

lib/src/expression/micro.dart
```diff
@@ -0,0 +1,111 @@
...
+  String directive, {
+  bool deSugarPipes: true,
+  String sourceUrl: '/test.dart',
+}) {
+  final assignments = <ReferenceAst>[];
+  final properties = <PropertyAst>[];
+  final parts = expression.split(';');
+
+  // Parse the first statement, which is special.
+  if (!parts[0].startsWith('let')) {
```
MichaelRFairhurst, line 37: What about ngIf? doesn't have to start with let

lib/src/expression/micro.dart
```diff
@@ -0,0 +1,111 @@
...
+  }
+  var parse = parts[0].split(' ');
+  assignments.add(new ReferenceAst(parse[1]));
+  var name = parse[2];
+  name = directive + name[0].toUpperCase() + name.substring(1);
+  properties.add(
+    new PropertyAst(
+      name,
+      new ExpressionAst.parse(
+        parse[3],
```
MichaelRFairhurst, line 48: could contain spaces, ie `let x of myMethod(arg1, arg2)`

Also, see work here that @mk13 has done related to trailing expressions, which you'll need to check: https://github.com/dart-lang/angular_analyzer_plugin/pull/94/files#diff-dd70e04facf5d905f5ff348ad54053f3R987

lib/src/expression/micro.dart
```diff
@@ -0,0 +1,111 @@
...
+
+  // Parse remaining statements, if any.
+  for (var part in parts.skip(1)) {
+    part = part.trim();
+    if (part.startsWith('let')) {
+      var name = part.substring(4, part.indexOf('='));
+      var value = part.substring(part.indexOf('=') + 1);
+      assignments.add(new ReferenceAst(name.trim(), value.trim()));
+    } else {
+      var subParts = part.split(':');
```
MichaelRFairhurst, line 63: I think `:` is allowed more than just here, ie `let x of: items`.

Also, I think `let x of items` is equivalent to `let x; of items`

In general I think the grammar is pretty simple:

```
Star := '*' name '=' '"' (expr | MicroExprList) '"'
MicroExprList = (LetExpr | BindExpr)+
LetExpr := 'let' id ('=' id)? ';'?
BindExpr := (id ':'?)? expr ';'?
```

Even simpler when you think of `*x="y"` as sugar for `template="x y"`

```
Template := 'template' '=' '"' MicroExprList '"'
```

lib/src/visitor.dart
```diff
@@ -0,0 +1,96 @@
...
+///
+/// An implementation may return element [R], and optionally use [C] as context.
+abstract class TemplateAstVisitor<R, C> {
+  const TemplateAstVisitor();
+
+  /// Visits an [astNode] tree, passing along [context] if given.
+  ///
+  /// Based on the subtype other `visit` methods are called.
+  R visit(TemplateAst astNode, [C context]) {
+    if (astNode is AttributeAst) {
```
MichaelRFairhurst, line 19: This should just call astNode.accept(this, context), no?

MichaelRFairhurst (approved):
Aside from trailing expressions and some lack of a spec to conform to in terms of how exactly star syntax works, this looks good.

lib/src/ast/attribute.dart
matanlurey: Nope, will fix in #14.

lib/src/ast/expression.dart
matanlurey: Ah good idea, added to #14.

lib/src/expression/micro.dart
matanlurey: AFAIK ngIf is always bound to straight expression, right?

lib/src/expression/micro.dart
matanlurey: Thanks! Added to #14

lib/src/expression/micro.dart
MichaelRFairhurst: Ah, I see, yeah, you check that later.

I think still technically `*ngFor="of items; let item"` is valid

MichaelRFairhurst: But could be wrong

lib/src/expression/micro.dart
matanlurey: Oh man, I had no idea that's what this even meant...

Then I need a much better micro expression parser. See #15

lib/src/visitor.dart
matanlurey: Yep, that didn't exist before. Filed in #14.

Issue #16, opened 2016-12-24
Figure out how to extend the AST to support Pipes better


matanlurey:
Potentially add a `PipeExpression` type so a `visitPipe` can be added for tooling.

@MichaelRFairhurst on https://github.com/dart-lang/angular_ast/pull/11:
> Hmm, I think the only thing about that to me is that the resulting AST has a semantic meaning, just one that existing tools won't like.
> 
> I think it'd be best to transform the semantic meaning into a valid one (ie, actually expose $$ng.pipe to users and everything), or create an extension to the AST so that existing tools know that they need to be extended with `visitPipe` etc before they can work.

Issue #15, opened 2016-12-24
Add a formal parser for * micro expressions


matanlurey:
From https://github.com/dart-lang/angular_ast/pull/12#discussion_r93811712:

> I think `:` is allowed more than just here, ie `let x of: items`.
> 
> Also, I think `let x of items` is equivalent to `let x; of items`
> 
> In general I think the grammar is pretty simple:
> 
> ```
> Star := '*' name '=' '"' (expr | MicroExprList) '"'
> MicroExprList = (LetExpr | BindExpr)+
> LetExpr := 'let' id ('=' id)? ';'?
> BindExpr := (id ':'?)? expr ';'?
> ```
> 
> Even simpler when you think of `*x="y"` as sugar for `template="x y"`
> 
> ```
> Template := 'template' '=' '"' MicroExprList '"'
> ```

MichaelRFairhurst:
this is BNF by the way (kind of like regexps)

Issue #14, opened 2016-12-24
Pre 0.2.0 polish required


matanlurey:
- [x] Remove `// TODO: Support pipe arguments.`
- [x] Ensure that the parser only accepts identifiers for pipe names
- [x] Use `PrefixedIdentifier` instead of `SimpleIdentifier`

~~Support parsing expressions without throwing when a failure occurs~~
~~Include offset data in `parseExpression`~~

- [x] Remove the hacky `/test.dart`
- [x] Import the analyzer `Expression` class prefixed

~~Remove `C` from `/*<R, C>*/` on `TemplateAst#accept` methods~~

Also see #15, and https://github.com/dart-lang/angular_ast/issues/17

Issue #17, opened 2016-12-25
Meta: Requirements use by pkg/angular_analyzer


matanlurey:
- [ ] Support parsing expressions without throwing when a failure occurs
- [ ] Include offset data in `parseExpression`
- [ ] Add `visitExpression` to `TemplateAstVisitor`
- [ ] Add support for `accept` to `BananaAst` and `StarAst` for tooling teams
- [ ] Check trailing expressions ([see implementation][trailing])
- [x] `TemplateAstVisitor#visit` should delegate to `TemplateAst#accept`.

[trailing]: https://github.com/dart-lang/angular_analyzer_plugin/pull/94/files#diff-dd70e04facf5d905f5ff348ad54053f3R987

PR #18, opened 2016-12-25
A bunch of cleanup work from prior PRs

matanlurey:
Closes https://github.com/dart-lang/angular_ast/issues/14
Closes https://github.com/dart-lang/angular_ast/issues/15
Closes https://github.com/dart-lang/angular_ast/issues/16

Moved some of the issues around making it more accessible for tooling into https://github.com/dart-lang/angular_ast/issues/17, which could use your input, but I didn't want to block this PR on it. I also realize the micro-expression parser is a bit of a mess - but it was definitely a simple improvement over the prior naive implementation.


matanlurey:
Ping @MichaelRFairhurst @jonahwilliams 

*/edit* TBR.

lib/src/expression/micro/parser.dart
```diff
@@ -0,0 +1,126 @@
...
+
+  void _parseBind() {
+    final name = _tokens.current.lexeme;
+    if (!_tokens.moveNext() ||
+        _tokens.current.type != NgMicroTokenType.bindExpressionBefore ||
+        !_tokens.moveNext() ||
+        _tokens.current.type != NgMicroTokenType.bindExpression) {
+      throw _unexpected();
+    }
+    final value = _tokens.current.lexeme;
```
MichaelRFairhurst, line 71: This can be a full expression, not just a lexeme, ie `let x of items.where(filter)`. I think you want to use the dart lexer instead of a custom one

lib/src/expression/micro/parser.dart
```diff
@@ -0,0 +1,126 @@
...
+    ));
+  }
+
+  void _parseLet() {
+    String identifier;
+    if (!_tokens.moveNext() ||
+        _tokens.current.type != NgMicroTokenType.letKeywordAfter ||
+        !_tokens.moveNext() ||
+        _tokens.current.type != NgMicroTokenType.letIdentifier) {
+      throw _unexpected();
```
MichaelRFairhurst, line 87: Once again, we may not want errors to throw if we want to analyze half complete templates

lib/src/expression/pipe.dart
```diff
@@ -0,0 +1,106 @@
...
+  String toSource() {
+    var parameterCall = parameters.map((p) => p.toSource()).join(':');
+    if (parameterCall.isNotEmpty) {
+      parameterCall = ':$parameterCall';
+    }
+    return '${value.toSource()} | ${name.name}$parameterCall';
+  }
+
+  @override
+  /*=E*/ accept/*<E>*/(AstVisitor/*<E>*/ visitor) {
```
MichaelRFairhurst, line 55: I think you should create an `NgAstVisitor` class which extends `AstVisitor` and has a `visitPipeExpression` method. You can then check `if (visitor is NgAstVisitor)` and throw something if its not, since the visitor might not work properly on the AST.

Although, maybe throwing isn't the right behavior...Many visitors would probably work on pipes just fine, and this is an example of how it might work well. However, you could make a mixin that adds `visitPipeExpression(x) => visitBinaryExpression(x)` to be easily used. That would at least work until someone makes a visitor which creates a visitor, in which case you can't easily add `PipeAsBinExpMixin` to the inner visitor without overriding the outer visitor when it creates the inner one....

Just some thoughts.

lib/src/expression/pipe.dart
```diff
@@ -0,0 +1,106 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/syntactic_entity.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/src/dart/ast/ast.dart';
+
+/// An implementation of [Expression] for the pipe operator in AngularDart.
+class PipeExpression extends ExpressionImpl implements BinaryExpression {
```
MichaelRFairhurst, line 12: I don't think this is really a binary expression. I think binary expressions imply that they have an expression on the left and an expression on the right.

As an example, `a.b` is not a binary expression, even though it has `a` as a LHS, `b` as a RHS, and `.` as an operator with some precedence. But its not a binary expression because the RHS is not an expression (the LHS still is though). I think that pretty closely matches what pipes are.

Not sure what the implications of using BinaryExpression are here though, might be harmless.

lib/src/expression/pipe.dart
```diff
@@ -0,0 +1,106 @@
...
+  @override
+  set leftOperand(_) {
+    throw new UnsupportedError('Cannot be modified');
+  }
+
+  @override
+  MethodElement propagatedElement;
+
+  @override
+  Expression get rightOperand => name;
```
MichaelRFairhurst, line 89: Yeah, `name` is `Identifier`, and `Identifier` may be an `Expression`, but its not really an expression, its really just an identifier

test/expression/micro/parser_test.dart
```diff
@@ -0,0 +1,70 @@
...
+          new ReferenceAst('foo'),
+        ],
+        properties: [],
+      ),
+    );
+  });
+
+  test('should parse a let assignment', () {
+    expect(
+      parse('ngThing', 'let foo = bar;let baz'),
```
MichaelRFairhurst, line 33: By the way, I managed to get this test case working. Don't think we need to support it as far as I'm aware, but its technically currently valid due to the whole `*x="y"` being equivalent to `template="x y"` thing.

template:
```
<div *templateComp="'expression'; plus: 'second expression'; let templateComp">
    &lt;div *templateComp="{{templateComp.templateComp}}; plus: {{templateComp.templateCompPlus}}; let templateComp"&gt; worked
</div>
```

output:
```
<div *templateComp="expression; plus: second expression; let templateComp"> worked
```

(technically it showed that twice for some bug but that's unimportant)

dart:
```
@Directive(selector: '[templateComp]', exportAs: 'templateComp')
class TemplateComponent {

  String _templateComp;
  String _templateCompPlus;
  TemplateRef _templateCompTemplate;
  @Input() set templateComp(String x) { _templateComp = x; rerender(); }
  @Input() set templateCompPlus(String x) { _templateCompPlus = x; rerender(); }
  @Input() set templateCompTemplate(TemplateRef x) { _templateCompTemplate = x; rerender(); }
  String get templateComp => _templateComp;
  String get templateCompPlus => _templateCompPlus;

  ViewContainerRef _viewContainer;

  TemplateComponent(this._viewContainer, this._templateCompTemplate);

  void rerender() {
    var view = this._viewContainer.createEmbeddedView(_templateCompTemplate);
    view.setLocal(r'$implicit', this);
    _viewContainer.insert(view);
  }

}
```

MichaelRFairhurst (commented):
Sorry to get this to you late

test/expression/micro/parser_test.dart
matanlurey: Cool thanks.

lib/src/expression/pipe.dart
matanlurey: I don't want to introduce type overrides here - I already have the correct types `name` and `value` above. In a future PR I could do something like:

```dart
abstract class PipeExpression implements Expression {
  Identifier get name;
  Expression get value;
}
```

And hide a bit of the binary expression-ish in a private class. Probably not worth yet.

lib/src/expression/pipe.dart
matanlurey: Let's keep as-is for now. Willing to change when we have a good reason to.

lib/src/expression/pipe.dart
matanlurey: This makes it easier, in that I can take an existing BInaryExpression and emit something that matches that interface. Unless you're planning on evaluating the expression statically I think keeping this for now is safe.

lib/src/expression/micro/parser.dart
matanlurey: Let's chat next week - I'll leave this as-is for now, and let's come up with a strategy that will allow format exceptions to be reported without blocking analysis or tooling.

lib/src/expression/micro/parser.dart
matanlurey: This is supported, the next line parses it using the full parser.

PR #19, opened 2017-01-06
Some cleanup based on tooling requirements

matanlurey:


MichaelRFairhurst (approved):
LGTM

test/expression/micro/parser_test.dart
```diff
@@ -41,6 +41,26 @@ void main() {
     );
   });
 
+  test('should parse a let with a full Dart expression', () {
+    expect(
+      parse('ngFor', 'let x of items.where(filter)'),
```
MichaelRFairhurst, line 46: Ah, but I think _here_ is where I got you in terms of wanting to use the dart tokenizer!

```
  let x of [";", ",", "."]
```

I think that will fail

MichaelRFairhurst: To clarify what you probably want to change it to: Once you hit a dart expression, you probably want to tokenize everything left in the template binding (in this case everything after `of`), then call `parseExpression` on that. The resulting `Expression` object will have an `endToken`. You want to resume your own parsing after `expression.endToken.next` and assert that the next char is `;`. That's because `parseExpression` inherently does not expect to use the whole token stream, it expects its part of a larger program.

That's what makes using your own tokenizer a bit of a tricky business. What we do is tokenize with the dart tokenizer only and we just check `if (token is KeywordToken && token.lexeme == 'let')` etc. Though that's not always super clean code so maybe there's a way to blend two tokenizers that works a bit better.

PR #22, opened 2017-01-23
Simpler tokenizer

mk13:
Groundwork for discussed simpler tokenizer that is agnostic of current html/angular state as discussed in 'Simpler Lexer Proposal'. It has minimal states just for better tokenizing results, but none that determines the validity of an html tag, comment, or text.

This can be easily merged into the existing scanner implementation by parsing through tokens instead of raw string. The existing scanner can be kept as a fail-fast version. A different version can be implemented that will not fail-fast but attempt to recover and inject synthetic nodes while also reporting errors.

### DIFF
--- a/lib/src/simple_token.dart
+++ b/lib/src/simple_token.dart
@@ -5,6 +5,10 @@ import 'package:quiver/core.dart';
 part 'token/simple_type.dart';
 
 class NgSimpleToken {
+  factory NgSimpleToken.bang(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.bang, offset);
+  }
+
   //Probably don't need
   factory NgSimpleToken.closeBrace(int offset) {
     return new NgSimpleToken._(NgSimpleTokenType.closeBrace, offset);
@@ -18,16 +22,17 @@ class NgSimpleToken {
     return new NgSimpleToken._(NgSimpleTokenType.closeParen, offset);
   }
 
-  factory NgSimpleToken.commentBegin(int offset) {
-    return new NgSimpleToken._(NgSimpleTokenType.commentBegin, offset);
-  }
-
-  factory NgSimpleToken.commentEnd(int offset) {
-    return new NgSimpleToken._(NgSimpleTokenType.commentEnd, offset);
+  factory NgSimpleToken.dash(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.dash, offset);
   }
 
   factory NgSimpleToken.doubleQuote(int offset) {
     return new NgSimpleToken._(NgSimpleTokenType.doubleQuote, offset);
+  }
+
+  factory NgSimpleToken.doubleQuotedText(int offset, String lexeme) {
+    return new NgSimpleToken(
+        new NgSimpleTokenType.doubleQuotedText(lexeme), offset);
   }
 
   factory NgSimpleToken.elementStart(int offset) {
@@ -36,10 +41,6 @@ class NgSimpleToken {
 
   factory NgSimpleToken.elementEnd(int offset) {
     return new NgSimpleToken._(NgSimpleTokenType.elementEnd, offset);
-  }
-
-  factory NgSimpleToken.elementEndVoid(int offset) {
-    return new NgSimpleToken._(NgSimpleTokenType.elementEndVoid, offset);
   }
 
   factory NgSimpleToken.EOF(int offset) {
@@ -75,8 +76,13 @@ class NgSimpleToken {
     return new NgSimpleToken._(NgSimpleTokenType.singleQuote, offset);
   }
 
+  factory NgSimpleToken.singleQuotedText(int offset, String lexeme) {
+    return new NgSimpleToken(
+        new NgSimpleTokenType.singleQuotedText(lexeme), offset);
+  }
+
   factory NgSimpleToken.text(int offset, String lexeme) {
-    return new NgSimpleToken(new NgSimpleTokenType.textType(lexeme), offset);
+    return new NgSimpleToken(new NgSimpleTokenType.text(lexeme), offset);
   }
 
   factory NgSimpleToken.unexpectedChar(int offset, String lexeme) {
@@ -85,8 +91,7 @@ class NgSimpleToken {
   }
 
   factory NgSimpleToken.whitespace(int offset, String lexeme) {
-    return new NgSimpleToken(
-        new NgSimpleTokenType.whitespaceType(lexeme), offset);
+    return new NgSimpleToken(new NgSimpleTokenType.whitespace(lexeme), offset);
   }
 
   const NgSimpleToken._(this.type, this.offset);
--- a/lib/src/simple_token.dart
+++ b/lib/src/simple_token.dart
@@ -43,15 +43,6 @@ class NgSimpleToken {
         NgSimpleTokenType.dashedIdentifier, offset, lexeme);
   }
 
-  factory NgSimpleToken.doubleQuote(int offset) {
-    return new NgSimpleToken._(NgSimpleTokenType.doubleQuote, offset, '"');
-  }
-
-  factory NgSimpleToken.doubleQuotedText(int offset, String lexeme) {
-    return new NgSimpleToken(
-        NgSimpleTokenType.doubleQuotedText, offset, lexeme);
-  }
-
   factory NgSimpleToken.tagStart(int offset) {
     return new NgSimpleToken._(NgSimpleTokenType.tagStart, offset, '<');
   }
@@ -97,15 +88,6 @@ class NgSimpleToken {
     return new NgSimpleToken._(NgSimpleTokenType.period, offset, '.');
   }
 
-  factory NgSimpleToken.singleQuote(int offset) {
-    return new NgSimpleToken._(NgSimpleTokenType.singleQuote, offset, "'");
-  }
-
-  factory NgSimpleToken.singleQuotedText(int offset, String lexeme) {
-    return new NgSimpleToken(
-        NgSimpleTokenType.singleQuotedText, offset, lexeme);
-  }
-
   factory NgSimpleToken.star(int offset) {
     return new NgSimpleToken._(NgSimpleTokenType.star, offset, '*');
   }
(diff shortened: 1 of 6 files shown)


lib/src/simple_token.dart
```diff
@@ -0,0 +1,133 @@
+library angular_ast.src.simple_token;
```
matanlurey, line 5: Copyright/License META for 2017

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,197 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
matanlurey: 2017

lib/src/token/simple_type.dart
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
matanlurey: 2017

test/simple_scanner_test.dart
```diff
@@ -0,0 +1,104 @@
+import 'package:angular_ast/src/simple_tokenizer.dart';
```
matanlurey, line 5: Copyright/License META for 2017

test/simple_tokenizer_test.dart
```diff
@@ -0,0 +1,269 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
matanlurey: 2017

test/simple_tokenizer_test.dart
```diff
@@ -0,0 +1,269 @@
...
+import 'package:test/test.dart';
+
+void main() {
+  Iterable<NgSimpleToken> tokenize(String html) =>
+      new NgSimpleTokenizer(html).tokenize();
+  String untokenize(Iterable<NgSimpleToken> tokens) => tokens
+      .fold(new StringBuffer(), (buffer, token) => buffer..write(token.lexeme))
+      .toString();
+
+  /**
```
matanlurey: ?

lib/src/simple_token.dart
```diff
@@ -0,0 +1,137 @@
...
+
+  factory NgSimpleToken.openParen(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.openParen, offset);
+  }
+
+  factory NgSimpleToken.singleQuote(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.singleQuote, offset);
+  }
+
+  factory NgSimpleToken.singleQuotedText(int offset, String lexeme) {
```
MichaelRFairhurst: Quotes could probably be an extension of `NgSimpleToken`. It can contain `openQuoteOffset` and `closedQuoteOffset` and even `isDoubleQuote`. That way we don't need to track six types of tokens (quote, closed quote, and unclosed quote all in single & double varieties) but we still don't throw out any information.

lib/src/simple_token.dart
```diff
@@ -0,0 +1,137 @@
...
+    return new NgSimpleToken._(NgSimpleTokenType.dash, offset);
+  }
+
+  factory NgSimpleToken.doubleQuote(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.doubleQuote, offset);
+  }
+
+  factory NgSimpleToken.doubleQuotedText(int offset, String lexeme) {
+    return new NgSimpleToken(
+        new NgSimpleTokenType.doubleQuotedText(lexeme), offset);
```
MichaelRFairhurst: I don't think the lexeme is part of the token _type_.

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,200 @@
...
+      token = scanner.scan();
+    }
+  }
+}
+
+class NgSimpleScanner {
+  String get state =>
+      _state == _NgSimpleScannerState.element ? "element" : "text";
+  String get remainingString => _scanner.rest;
+  String get elementRegex => _allElementMatches.toString();
```
MichaelRFairhurst: Looks like this can be removed?

lib/src/simple_token.dart
```diff
@@ -0,0 +1,137 @@
...
+  factory NgSimpleToken.singleQuotedText(int offset, String lexeme) {
+    return new NgSimpleToken(
+        new NgSimpleTokenType.singleQuotedText(lexeme), offset);
+  }
+
+  factory NgSimpleToken.star(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.star, offset);
+  }
+
+  factory NgSimpleToken.text(int offset, String lexeme) {
```
MichaelRFairhurst, line 95: Its customary to split up text types. Here we probably want text and identifier to be treated separately I would think.

lib/src/simple_token.dart
```diff
@@ -0,0 +1,137 @@
...
+
+  factory NgSimpleToken.commentBegin(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.commentBegin, offset);
+  }
+
+  factory NgSimpleToken.commentEnd(int offset) {
+    return new NgSimpleToken._(NgSimpleTokenType.commentEnd, offset);
+  }
+
+  factory NgSimpleToken.dash(int offset) {
```
MichaelRFairhurst, line 37: This should probably not be a token.

Better idea would be to make a `dashedIdentifier` token, `[a-Z0-9\-]+`, then in parseTag we can do things like `if (token is DashedIdentifier || token is Identifier)` and continue, while things like parseOutputBinding can check `if (token is Identifier)` thus reporting parse errors for `(a-b)` but not `<a-b>`

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,200 @@
...
+  static final _allElementMatches = new RegExp(r'(\])|' //1  ]
+      r'(\!)|' //2  !
+      r'(\-)|' //3  -
+      r'(\))|' //4  )
+      r'(>)|' //5  >
+      r'(\/)|' //6  /
+      r'(\[)|' //7  [
+      r'(\()|' //8  (
+      r'([\s]+)|' //9 whitespace
+      r'([a-zA-Z0-9]([a-zA-Z0-9\-\_]*[a-zA-Z0-9])?)|' //10 any alphanumeric + '-' + '_'
```
MichaelRFairhurst: Any reason why this excludes `-abc`?

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,200 @@
...
+      r'("([^"\\]|\\.)*")|' //12 closed double quote (includes group 13)
+      r"('([^'\\]|\\.)*')|" //14 closed single quote (includes group 15)
```
MichaelRFairhurst: If matched/unmatched/etc are not distinguished by token type, you can do `r("([^"\\]|\\.)*"?` to include unmatched tags in this one pattern

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,200 @@
...
+      case _NgSimpleScannerState.text:
+        return scanText();
+    }
+    return null;
+  }
+
+  NgSimpleToken scanComment() {
+    int offset = _scanner.position;
+    while (true) {
+      if (_scanner.peekChar() == $dash &&
```
MichaelRFairhurst, line 80: should this be a regex?

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,200 @@
...
+    if (_scanner.peekChar() == null || _scanner.rest.length == 0) {
+      return new NgSimpleToken.EOF(offset);
+    }
+    if (_scanner.scan(_allTextMatches)) {
+      Match match = _scanner.lastMatch;
+      if (matchesGroup(match, 1)) {
+        return new NgSimpleToken.text(offset, _scanner.substring(offset));
+      }
+      if (matchesGroup(match, 2)) {
+        _state = _NgSimpleScannerState.comment;
```
MichaelRFairhurst, line 188: Probably worth scanning the whole comment here rather than splitting it up into "open comment, comment text, close comment." Will make it easier in the body of the parser to step past comments if it can always just say "next()" rather than having to go past up to three times and check for EOF along the way.

Errors here (such as unclosed comments) should be reported here rather than in the parser for the same reason

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,200 @@
...
+        return new NgSimpleToken.commentBegin(offset);
+      }
+      if (matchesGroup(match, 3)) {
+        _state = _NgSimpleScannerState.element;
+        return new NgSimpleToken.tagStart(offset);
+      }
+    }
+    return new NgSimpleToken.unexpectedChar(
```
MichaelRFairhurst, line 196: Might be better to report the error internally and move past it rather than return it

lib/src/simple_tokenizer.dart
```diff
@@ -0,0 +1,200 @@
...
+        return new NgSimpleToken.whitespace(offset, _scanner.substring(offset));
+      }
+      if (matchesGroup(match, 10)) {
+        return new NgSimpleToken.text(offset, _scanner.substring(offset));
+      }
+      if (matchesGroup(match, 11)) {
+        return new NgSimpleToken.text(offset, _scanner.substring(offset));
+      }
+      if (matchesGroup(match, 12)) {
+        return new NgSimpleToken.doubleQuotedText(
```
MichaelRFairhurst: This needs to be de-escaped, ie: `<div [hidden]="something.contains(\"this\")"`

MichaelRFairhurst (approved):
Could be tweaked in a few ways, but this is a HUGE win for error recoverability.

lib/src/simple_token.dart
mk13: That's a good idea. I primarily kept dash as a separate token so that we can spot dangling dashes outside of identifier scope. 

For example: `<supposed to be a comment  -->`. This will tokenize the two dashes, but I suppose the unexpected character token can catch this.

lib/src/simple_token.dart
mk13: Sure, I can move the lexeme to be part of the Token instead of the type.

lib/src/simple_token.dart
mk13: Sure. I'll also ensure that the extension includes what type of quote it is (single/double).

lib/src/simple_token.dart
mk13: I wanted to keep this tokenizer to be agnostic about what can be considered a text node block or an identifier and let the parser that utilizes this tokenizer decide on how to treat the text. I've instead created different text regexp for text, comment, and identifier just so text blocks with spaces won't be treated as multiple redundant texts. 

We can further discuss what the best course of action for this is; I feel like if we classify texts as identifier, textnode text, or comment - we wouldn't be deviating much from the initial scanner/lexer implemented.

lib/src/simple_tokenizer.dart
mk13: Oh yes, that was there for debugging purposes.

lib/src/simple_tokenizer.dart
mk13: A tag that starts with a dash doesn't seem to be valid when I tried. For example `<-abc></-abc>`, or am I thinking of something else?

lib/src/simple_tokenizer.dart
mk13: I've initially done this as a regex, but there were too many edge cases where the regex would fail. A regex that worked would force me to return comment texts into multiple token instead of one single text token; this method seemed to yield the best accurate results.

lib/src/simple_tokenizer.dart
mk13: The current tokenizer should be returning unmatched quote tokens separately from a matched/closed quote block. 

`"some text` -> yields DoubleQuote token and texts tokens.
`"some text"` -> yields a single DoubleQuotedText token.

Cases where dangling double quote or single quotes are found, it is an easy indicated of an error.

lib/src/simple_tokenizer.dart
mk13: Sure, will de-escape this.

lib/src/simple_tokenizer.dart
mk13: I was thinking of keeping the simple tokenizer completely error-agnostic and let the parser that utilizes this tokenizer decide and call all errors. This way we can keep the actual tokenizer as a static const entity. Let's discuss tomorrow if this is the best course of action or to let the tokenizer immediately report unexpectedChars.

lib/src/simple_token.dart
MichaelRFairhurst: Ah, I think in this case that could be treated as an error token

lib/src/simple_token.dart
MichaelRFairhurst: I don't think the tokenizer should be agnostic. A text block will not occur inside a property, and neither will a comment. Moreover, they have different characteristics (a comment block has open & close delimiters, a nodename doesn't have spaces, an attribute name doesn't have dashes...). Its customary to split different types of identifiers...even "ng-content", "select", "on", "bind" etc. could be treated as its own angular "keyword"s so that the parser doesn't accidentally mix them with other non-special identifiers.

The main point here is to let the typesystem catch mistakes in the parser. "token is X" should rule out all garbage possibilities, such as a text node erroneously inside a tag, or a "bind" keyword inside a "[...]" binding. etc etc

lib/src/simple_tokenizer.dart
MichaelRFairhurst: Yeah, I guess that makes sense

lib/src/simple_tokenizer.dart
MichaelRFairhurst: I think

`"some text` -> should yield a `StringToken` where type is `double`, `endQuoteOffset` is null and contents are `some text`
`"some text"` -> should yield a `StringToken` where type is `double`, `endQuoteOffset` is not null, and contents are `some text`

If we do something fancy like trying to recover from `<x y="z q="w"`, then we should do that in an experimental branch once we see a need for it. And in theory, as long as we track that the end quote is missing, we should be able to recover in either case.

lib/src/simple_tokenizer.dart
MichaelRFairhurst: Could be split up into two iterators. The resulting iterators don't have to be const

MichaelRFairhurst:
recent changes look good! Still approved

matanlurey (approved):

PR #23, opened 2017-02-07
Implement simple tokenizer

mk13:
Submitting a PR as an intermediate checkpoint to get some thoughts and opinions before continuing further.

Primary changes: 
- Parsed AST nodes are explicit types that will provide more information. Only parsed/real AST nodes are obligated to preserve offsets; synthetics from the result of desugaring do not. 
- Desugaring is now a separate process that doesn't occur automatically with the parsing. A Desugaring visitor is used to desugar banana and star syntax. This gives users of the tool more control. This is reflected upon the primary top level 'parse' function with flags to include .
- Tokens are condensed into two main files: types and tokens.
- Wrapper Token types added for attribute name and attribute value to be used in the different AST nodes. 

Things to do and concerns: 
- Error Listener and accumulation (along with recovery) is being worked on in another branch.
- Stronger, more universal 'API' for the various AST nodes. There is still a lot of ad-hoc variables that can be cleaned out and generalizes across the ast nodes. 
- Should parsed ng-content and templates allow for trailing whitespace within the opener tag? For example <ng-content select="blah"    >  or <template   >? It is currently being allowed for normal elements. I feel like it should be strictly universal across these three types.
- Should we allow for Token exposure for developers utilizing this tool? Or should we try to keep the Tokens completely hidden and abstracted away? I can see arguments for both; developers who need explicit control of various components can utilize the Tokens and we don't have to juggle with as much 'getters' within the AST node. On the other hand, we're increasing the complexity of the tool.


lib/angular_ast.dart
```diff
@@ -16,30 +16,41 @@ export 'package:angular_ast/src/ast.dart'
...
 /// Returns [template] parsed as an abstract syntax tree.
-List<TemplateAst> parse(
-  String template, {
-  @required String sourceUrl,
-  bool toolFriendlyAst: false,
-}) {
+List<TemplateAst> parse(String template,
+    {@required String sourceUrl,
+    bool toolFriendlyAst: false,
+    bool desugar: false}) {
```
matanlurey: Can you document in `parse` what `desugar` means.

We might also want to default this to `true`, as I imagine most end users want to consume, not analyze, the AST.

lib/src/ast/attribute.dart
```diff
@@ -13,10 +13,7 @@ import 'package:quiver/core.dart';
 /// Clients should not extend, implement, or mix-in this class.
 abstract class AttributeAst implements TemplateAst {
   /// Create a new synthetic [AttributeAst] with a string [value].
-  factory AttributeAst(
-    String name, [
-    String value,
-  ]) = _SyntheticAttributeAst;
+  factory AttributeAst(String name, [String value]) = _SyntheticAttributeAst;
```
matanlurey: This is a relatively small formatting point, but do you mind reverting the changes removing all the trailing commas here and across the PR? I realize it's overkill for 1-2 parameters, but because there are some classes with many, I'd prefer them _all_ to have trailing commas for consistency.

I imagine it also will make the delta of this CL much smaller.

lib/src/ast/attribute.dart
```diff
@@ -55,33 +50,51 @@ abstract class AttributeAst implements TemplateAst {
...
+      return '$AttributeAst {$name=$quotedValue}';
     }
     return '$AttributeAst {$name}';
   }
 }
 
-class _ParsedAttributeAst extends TemplateAst with AttributeAst {
-  final NgToken _nameToken;
-  final NgToken _valueToken;
+class ParsedAttributeAst extends TemplateAst with AttributeAst, OffsetInfo {
```
matanlurey: Here and elsewhere in the PR - can we get a 1-line document for all the ParsedXXX clases, and their public properties (like `nameToken`, below)?

I want to make sure the generated dartdoc isn't too mysterious for new users - even if it's obvious enough to us.

lib/src/ast/element.dart
```diff
@@ -97,6 +101,19 @@ abstract class ElementAst implements StandaloneTemplateAst {
   /// Reference assignments.
   List<ReferenceAst> get references;
 
+  /// Bananas assignments.
+  List<BananaAst> get bananas;
+
+  ///Star assignments.
+  List<StarAst> get stars;
+
+  set bananas(List<BananaAst> l);
```
matanlurey: Why do we want public setters for bananas/stars for `ElementAst`?

lib/src/ast/element.dart
```diff
@@ -134,24 +169,30 @@ abstract class ElementAst implements StandaloneTemplateAst {
...
-    this.attributes: const [],
-    this.childNodes: const [],
-    this.events: const [],
-    this.properties: const [],
-    this.references: const [],
-  })
+class ParsedElementAst extends TemplateAst with ElementAst {
+  final NgToken identifierToken;
+  final int openTagEnd;
+  final int closeTagStart; //Can be null if void element
```
matanlurey: Comment like that should be public, i.e.

/// May be null if a void `<tag />` element (no closing tag).

matanlurey (commented):
I totally agree with how this was done, and generally LGTM - some small comments about consistency and documentation I'd like to see addressed before approval - but otherwise amazing job.

lib/src/ast/element.dart
mk13: Was used to empty out bananas and stars after desugaring occurs. Will fix to just clear().

mk13:
I'm currently working on a documentation to explain the design and objectives of the changes. Will be provided as soon as it's finished.

lib/angular_ast.dart
```diff
@@ -16,30 +16,41 @@ export 'package:angular_ast/src/ast.dart'
...
+        DesugarVisitor;
 
 /// Returns [template] parsed as an abstract syntax tree.
-List<TemplateAst> parse(
-  String template, {
-  @required String sourceUrl,
-  bool toolFriendlyAst: false,
-}) {
+List<TemplateAst> parse(String template,
+    {@required String sourceUrl,
```
liuming0: Since sourceUrl is required, is there a reason why you make it named parameter?

lib/src/ast/attribute.dart
```diff
@@ -55,33 +50,51 @@ abstract class AttributeAst implements TemplateAst {
   /// Static attribute value; may be `null` to have no value.
   String get value;
 
+  /// Static attribute value with quotes attached;
+  /// may be `null` to have no value.
+  String get quotedValue;
+
   @override
   String toString() {
     if (value != null) {
```
liuming0: Since you changed to quotedValue, do you need to change here too?

lib/src/ast/attribute.dart
```diff
@@ -91,6 +104,9 @@ class _SyntheticAttributeAst extends SyntheticTemplateAst with AttributeAst {
   @override
   final String value;
 
+  @override
+  String get quotedValue => '"$value"';
```
liuming0: check if value is null.

lib/src/ast/element.dart
```diff
@@ -97,6 +101,19 @@ abstract class ElementAst implements StandaloneTemplateAst {
...
+  List<BananaAst> get bananas;
+
+  ///Star assignments.
+  List<StarAst> get stars;
+
+  set bananas(List<BananaAst> l);
+
+  set stars(List<StarAst> l);
+
+  ///Whitespaces
```
liuming0: space after ///

lib/src/parser/reader.dart
```diff
@@ -50,8 +68,92 @@ class NgTokenReader {
...
+    return null;
+  }
 
   /// Returns whether the current token is of [type].
-  bool when(NgTokenType type) => _iterator.current.type == type;
+  bool when(NgBaseTokenType type) => _iterator.current.type == type;
+
+  /// Returns whether there is any more tokens to return.
+  bool get isDone {
+    if (_peek != null) {
```
liuming0: Do you mean "_peek == null"

lib/src/token/token_types.dart
```diff
@@ -0,0 +1,242 @@
...
+
+  static const closeParen = const NgSimpleTokenType._('closeParen');
+
+  static const commentBegin = const NgSimpleTokenType._('commentBegin');
+
+  static const commentEnd = const NgSimpleTokenType._('commentEnd');
+
+  static const dash = const NgSimpleTokenType._('dash');
+
+  //static const dashedIdentifier = const NgSimpleTokenType._('dashedIdentifier');
```
liuming0: delete?

lib/src/token/token_types.dart
```diff
@@ -0,0 +1,242 @@
...
+  String get name;
+}
+
+/// The types of tokens that can be returned by the NgStringTokenizer
+///
+/// Clients may not extend, implement, or mix-in this class.
+class NgSimpleTokenType implements NgBaseTokenType {
+  static const bang = const NgSimpleTokenType._('bang');
+
+  //probably not needed
```
liuming0: delete?

lib/src/token/token_types.dart
```diff
@@ -0,0 +1,242 @@
...
+
+  static const commentEnd = const NgSimpleTokenType._('commentEnd');
+
+  static const dash = const NgSimpleTokenType._('dash');
+
+  //static const dashedIdentifier = const NgSimpleTokenType._('dashedIdentifier');
+
+  static const doubleQuote = const NgSimpleTokenType._('doubleQuote');
+
+  static const tagStart = const NgSimpleTokenType._(
```
liuming0: I think this can be done in one line.

lib/src/token/token_types.dart
```diff
@@ -0,0 +1,242 @@
...
+
+  static const hash = const NgSimpleTokenType._('hash');
+
+  static const identifier = const NgSimpleTokenType._('identifier');
+
+  static const mustacheBegin = const NgSimpleTokenType._('mustacheBegin');
+
+  static const mustacheEnd = const NgSimpleTokenType._('mustacheEnd');
+
+  //Probably not needed
```
liuming0: delete?

lib/src/visitors/desugar_visitor.dart
```diff
@@ -0,0 +1,141 @@
...
+            bananaAst.accept(this, DesugarFlag.property);
+        TemplateAst toAddEvent = bananaAst.accept(this, DesugarFlag.event);
+        astNode.properties.add(toAddProperty);
+        astNode.events.add(toAddEvent);
+      }
+      astNode.bananas = const [];
+    }
+
+    if (astNode.stars.isNotEmpty) {
+      //TODO: Look into better structure for better AST visitor pattern
```
liuming0: space after //

lib/angular_ast.dart
mk13: Hmmm I just left that as it was from the original before I started working on it. I think the intent is to make users of this function to explicitly differentiate the template from the sourceUrl through named parameter.

mk13:
A lot of this code is outdated due to changes in error recovery; I'll likely create another PR entirely that includes changes within this PR.

kevmoo:
Yes – rebasing over latest should also get travis to a known good state.

mk13:
Refer to PR https://github.com/dart-lang/angular_ast/pull/25 ; contains these changes.

PR #25, opened 2017-02-23
Error recovery implementation

mk13:
Combines changes from PR: https://github.com/dart-lang/angular_ast/pull/23

In addition, added the following: 
-Error recovery at the micro level, not at the AST level. That will be a separate PR. 
-Refer to test/recover_errors_lexer_test.dart for explicit usecases of recovery.
-Added interactive cli tester (test/cli_tester.dart) for recovery at the lexer level. 

Future changes to expect:
-Ast level recovery
-Recovery with interpolations; not yet fully implemented yet. 
-More refactoring and code cleanup.
-TokenTypes will be changes to enums in a separate PR for better performance.

Still far from exiting development status, but this is a good checkpoint for lexer level error recovery.


matanlurey (approved):
New changes LGTM. Thanks for the hard work!

lib/src/ast/element.dart
```diff
@@ -97,6 +112,15 @@ abstract class ElementAst implements StandaloneTemplateAst {
...
   List<ReferenceAst> get references;
 
+  /// Bananas assignments.
+  List<BananaAst> get bananas;
+
+  /// Star assignments.
+  List<StarAst> get stars;
+
+  /// Whitespaces
+  List<WhitespaceAst> get whitespaces;
```
MichaelRFairhurst, line 122: Not sure we need this, we can figure out whitespace by looking at the gaps between the ranges that we _are_ recording. Unless we ever have two `WhitespaceAst`s next to each other and we need to preserve them

lib/src/ast/offset_info.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Mixin used to preserve offsets of tokens to be able to reproduce the same
+/// text. In addition, preserves offsets in cases where banana syntax and
+/// template syntax are desugared.
+abstract class OffsetInfo {
```
MichaelRFairhurst: Looks pretty specfiic to tags, maybe should be named as such too?

lib/src/ast/property.dart
```diff
@@ -98,34 +101,88 @@ abstract class PropertyAst implements TemplateAst {
...
+/// Represents a real, non-synthetic bound property assignment
+/// `[name.postfix.unit]="value"`for an element.
+///
+/// Clients should not extend, implement, or mix-in this class.
+class ParsedPropertyAst extends TemplateAst
+    with PropertyAst, OffsetInfo, SpecialOffsetInfo {
+  /// Tokens representing `[property]` attribute.
+  final NgToken prefixToken;
+  final NgToken elementDecoratorToken;
+  final NgToken suffixToken;
```
MichaelRFairhurst: Comparing these fields to the comment above, is this "unit"? and is "prefixToken" what the comment describes as "name"?

lib/src/ast/property.dart
```diff
@@ -98,34 +101,88 @@ abstract class PropertyAst implements TemplateAst {
...
-  }
+  String get _nameWithoutBrackets => elementDecoratorToken.lexeme;
 
+  /// Name `name` of `[name.postfix.unit]`.
   @override
   String get name => _nameWithoutBrackets.split('.').first;
 
+  /// Offset of name.
+  @override
+  int get nameOffset => elementDecoratorToken.offset;
```
MichaelRFairhurst: Does this need to use '?.offset'?

lib/src/ast/property.dart
```diff
@@ -98,34 +101,88 @@ abstract class PropertyAst implements TemplateAst {
...
+  @override
+  int get valueOffset => valueToken?.innerValue?.offset;
+
+  /// Offset of value starting at left quote; may be `null` to have no value.
+  @override
+  int get quotedValueOffset => valueToken?.leftQuote?.offset;
+
+  /// Offset of `[` prefix in `[name.postfix.unit]`.
+  @override
+  int get specialPrefixOffset => prefixToken.offset;
```
MichaelRFairhurst: why is this called "special"?

lib/src/ast/sugar/banana.dart
```diff
@@ -44,60 +47,107 @@ abstract class BananaAst implements TemplateAst {
   @override
   bool operator ==(Object o) {
     if (o is BananaAst) {
```
MichaelRFairhurst, line 49: what is this used for? would probably better to use your new fancy shiny offsets to make sure its the same one

lib/src/ast/whitespace.dart
```diff
@@ -0,0 +1,49 @@
...
+import 'package:angular_ast/src/ast.dart';
+import 'package:angular_ast/src/token/tokens.dart';
+import 'package:angular_ast/src/visitor.dart';
+import 'package:source_span/source_span.dart';
+
+/// Represents a generic-use whitespace ast to preserve offsets
+/// where needed. This cannot exist as a synthetic and is always real.
+///
+/// Clients should not extend, implement, or mix-in this class.
+class WhitespaceAst extends TemplateAst {
```
MichaelRFairhurst, line 14: Once again, not sure if this is necessary.

Can't we see when "<div" is at offsets 0-4, and that ">" is at offset 7, to figure out that there are three whitespace chars ie `<div   >`? Feasibly whitespace can be in so many places, it seems to me like we shouldn't try to track it specially in any of the places it could occur.

lib/src/exception_handler/exception_handler.dart
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+abstract class ExceptionHandler {
+  void handle(Exception e);
```
MichaelRFairhurst, line 8: I think exception classes are pretty heavyweight to _not_ throw, and also pretty broad.

I would think that this should be like "ParseErrorHandler" with a "ThrowingParseErrorHandler" which throws "new ParseAbortException(error)". Otherwise clients that want to just listen to parse errors have full exception objects at all times, which doesn't make sense if they aren't supposed to be thrown. Also for what its worth in dart you can throw anything, not just exceptions, so it could even be `String errMsg` and you can `throw errMsg`, so even for the fast-failure case, a full `Exception` isn't really necessary.

lib/src/parser.dart
```diff
@@ -45,19 +47,44 @@ class NgParser {
...
+    final tokens = const NgLexer().tokenize(template, exceptionHandler);
+    final parser = new RecursiveAstParser(
+      new SourceFile(
+        template,
+        url: sourceUrl,
+      ),
+      tokens,
+      _voidElements,
+    );
+    List<StandaloneTemplateAst> asts = parser.parse();
```
MichaelRFairhurst: looks like tokens and parser are final, ast and visitor could be final too. (and the code would be leaner, you can omit both types).

lib/src/parser.dart
```diff
@@ -45,19 +47,44 @@ class NgParser {
...
+    DesugarVisitor visitor =
+        new DesugarVisitor(toolFriendlyAstOrigin: _toolFriendlyAstOrigin);
+    return asts.map((t) => t.accept(visitor)).toList();
+  }
+
+  /// Return a series of tokens by incrementally scanning [template].
+  ///
+  /// Does not automatically desugar.
+  List<StandaloneTemplateAst> parsePreserve(
```
MichaelRFairhurst, line 70: Looks like `parse` can calll `parsePreserve` and then visit the result

lib/src/parser/reader.dart
```diff
@@ -50,8 +75,113 @@ class NgTokenReader {
...
+  NgBaseToken peek() => _peek = next();
+
+  /// Returns the next token type without incrementing.
+  /// Returns null otherwise.
+  NgBaseTokenType peekType() {
+    _peek = next();
+    if (_peek != null) {
+      return _peek.type;
+    }
+    return null;
```
MichaelRFairhurst, line 88: shouldn't this always return something, at least EOF?

lib/src/parser/reader.dart
```diff
@@ -50,8 +75,113 @@ class NgTokenReader {
...
+  }
 
   /// Returns whether the current token is of [type].
-  bool when(NgTokenType type) => _iterator.current.type == type;
+  bool when(NgBaseTokenType type) => _iterator.current.type == type;
+
+  /// Returns whether there is any more tokens to return.
+  bool get isDone {
+    if (_peek != null) {
```
MichaelRFairhurst, line 96: this should go off of EOF. avoids null pointer exceptions

lib/src/parser/reader.dart
```diff
@@ -50,8 +75,113 @@ class NgTokenReader {
...
+  )
+      : super._(source, iterator);
+
+  /// Returns the next token if it is of [type].
+  /// Otherwise throws a [FormatException].
+  /// Do not use where errorRecovery is potentially enabled.
+  @override
+  NgBaseToken expect(NgBaseTokenType type) {
+    final next = this.next();
+    String message = 'Expected a token of $type but got ${next.type}';
```
MichaelRFairhurst: caveat: string building can be expensive, this can be moved after the if statement.
caveat to the caveat: the JIT should do this for you eventually most likely after a certain point

lib/src/parser/reader.dart
```diff
@@ -50,8 +75,113 @@ class NgTokenReader {
...
+    error(message);
+    return null;
+  }
+
+  /// Scans forward for the next peek type that isn't ignoreType
+  /// For example, `peekTypeIgnoringType(whitespace)` will peek
+  /// for the next type that isn't whitespace.
+  /// Returns `null` if there are no further types aside from ignoreType
+  /// or iterator is empty.
+  NgBaseTokenType peekTypeIgnoringType(NgBaseTokenType ignoreType) {
```
MichaelRFairhurst, line 145: Is there any use-case for this except whitespace? Might be better to do "peekIgnoringWhitespace"

lib/src/parser/recursive.dart
```diff
@@ -55,67 +53,102 @@ class RecursiveAstParser {
...
-    // - attribute
-    switch (startChar) {
-      case $open_bracket:
-        return new PropertyAst.parsed(
+    NgAttributeValueToken valueToken;
+    NgToken equalSignToken;
+
+    if (_reader.peekTypeIgnoringType(NgTokenType.whitespace) ==
+        NgTokenType.beforeElementDecoratorValue) {
+      while (_reader.peekType() == NgTokenType.whitespace) {
```
MichaelRFairhurst, line 83: You added a method on reader to do this right? Would be good for using here too

lib/src/parser/recursive.dart
```diff
@@ -55,67 +53,102 @@ class RecursiveAstParser {
...
-    // - event
-    // - reference
-    // - attribute
-    switch (startChar) {
-      case $open_bracket:
-        return new PropertyAst.parsed(
+    NgAttributeValueToken valueToken;
+    NgToken equalSignToken;
+
+    if (_reader.peekTypeIgnoringType(NgTokenType.whitespace) ==
```
MichaelRFairhurst, line 81: Does this mean it will still accept `[foo] = "bar"`?

lib/src/parser/recursive.dart
```diff
@@ -124,6 +157,7 @@ class RecursiveAstParser {
   StandaloneTemplateAst parseElement(NgToken beginToken) {
     // Parse the element identifier.
     final nameToken = _reader.expect(NgTokenType.elementIdentifier);
+    //TODO: do we want to allow trailing spaces in content/template tags?
```
MichaelRFairhurst, line 160: Good question.

I think usually `<| foo` means they are filling in the tag name. But `< foo|` could mean the same thing. Or it could be a property named foo...but we couldn't autocomplete that without a context anyway. If we could make both work, that would be cool. At least, in theory it would be. Don't know if users would come up against that often

lib/src/recovery_protocol/angular_analyzer_protocol.dart
```diff
@@ -0,0 +1,602 @@
...
+      int _offset = (current is NgSimpleQuoteToken)
+          ? current.quoteOffset
+          : current.offset;
+      reader.putBack(current);
+      returnState = NgScannerState.scanElementDecoratorValue;
+      returnToken = new NgToken.generateErrorSynthetic(
+          _offset, NgTokenType.beforeElementDecoratorValue);
+    }
+
+    return new RecoverySolution(returnState, returnToken);
```
MichaelRFairhurst: This is a really interesting solution.

I'm gonna drop some theory on you!

So in the programming world, you have regular languages, and you have context-free grammars. The difference between them basically boils down to one using a finite amount of memory and the other not. Regular languages can be read by a state machine, like so:

```
vertical = state
horizontal = sees char
value = next state, E for error, A for accept (finish)

   a b c d
  +-+-+-+-+
1 |1|2|3|4|
2 |E|2|3|4|
3 |E|E|3|4|
4 |E|E|E|A|
```

This is what the regex `/a*b*c*e{2}/` describes. Regexes are named after the regular languages that they describe/match/validate.

But by having a finite number of states, they have a finite amount of memory. You can't match, for instance, matching close parenthesis in a regular language, because th nth open paren needs n states to track n close parens. But the nth open paren can be followed by the n+1th open paren, so no finite number of states can track an unlimited number of open/close parens. You have to have a maximum amount of open/close parens to track, ie `/open (open (open close)? close)? close/` can match up to three open/close parens. (Note: many regex engines support a few things that go beyond regular languages, I believe in perl for instance you can write a regex for matching parens, it just isn't really a regex but that only matters to theorists).

So on the other side you have context-free grammars, which are basically just regular languages with a stack of states. Since stacks are unbounded, matching close parens is:

```
   (      )      EOF
  +------+------+------|
1 |push 2|     3|     E|
2 |push 2|pop   |     E|
3 |     E|     E|     A|
```

In compilers, the parsers and lexers are split partly because tokenization is a regular language (some amount of alpha chars in a row), and parsing is a context free grammar (matching html tags).

A lexer which has states and uses regexes is really just a bigger regex, if you think about it. Take any regex which has a final "accept" state, and just turn it into a transition into the first state of regex of the tokenizer's next state. So if beforeFoo has some state machine that ends when state 4 sees an "f", and "atFoo" is a state machine, then you just have one state machine: when state 4 of beforeFoo sees an "f", it transitions to state 1 of atFoo.

I was very very very wary of adding error recovery at the tokenization level for this reason. And in fact, some error recovery (such as unmatched html tag closes) will not be possible to recover from at tokenization. My whole goal of the simple tokenizer was to make it possible for the parser to recover from errors, and not the tokenizer.

BUT!

Html (and angular) is very very very close to regular language. Unlike dart, if you were to hand write a context-free-grammar state machine to process angular, it would have very very very few push state/pop state actions. Basically push on <id and pop on </id>, where dart has like "push start class" "push start if" "push start for" "push start map literal" "push start generic arguments" "push start function parameters" "push start optional function parameters" etc etc etc etc.

So we can do almost all of our error recovery at the tokenization level, as your branch proves. And this way you can do this nicely modular error recovery strategy that we can fine tune. Maybe we'll add a dash of unmatched tag recovery, and BAM. We have everything we need! (and unmatched tag recovery might never work worth a damn anyways).

Which means, if you think about it...

You haven't just implemented error recovery for angular.

You've implemented error recovery for regular expressions themselves!!!!!

Very cool.

I wonder if there's a way to apply this to regex engines in general.

lib/src/recovery_protocol/angular_analyzer_protocol.dart
```diff
@@ -0,0 +1,602 @@
...
+        type == NgSimpleTokenType.commentBegin ||
+        type == NgSimpleTokenType.openTagStart ||
+        type == NgSimpleTokenType.closeTagStart ||
+        type == NgSimpleTokenType.EOF ||
+        type == NgSimpleTokenType.doubleQuote ||
+        type == NgSimpleTokenType.singleQuote) {
+      reader.putBack(current);
+      int _offset = (current is NgSimpleQuoteToken)
+          ? current.quoteOffset
+          : current.offset;
```
MichaelRFairhurst: maybe these would be more useful flipped -- offset and content offset?

lib/src/recovery_protocol/recovery_protocol.dart
```diff
@@ -0,0 +1,207 @@
...
+      case NgScannerState.scanSuffixBanana:
+        return scanSuffixBanana(current, reader);
+      case NgScannerState.scanSuffixEvent:
+        return scanSuffixEvent(current, reader);
+      case NgScannerState.scanSuffixProperty:
+        return scanSuffixProperty(current, reader);
+      case NgScannerState.scanText:
+        return scanText(current, reader);
+    }
+    return new RecoverySolution.skip();
```
MichaelRFairhurst, line 73: maybe this should throw instead?

lib/src/scanner.dart
```diff
@@ -2,326 +2,622 @@
...
-        _scanner.peekChar() == $slash) {
-      return scanElementEnd(wasOpenTag: true);
-    } else if (_scanner.matches(_findWhitespace)) {
+    if (_current.type == NgSimpleTokenType.equalSign) {
+      _state = NgScannerState.scanElementDecoratorValue;
+      return new NgToken.beforeElementDecoratorValue(_current.offset);
+    } else if (_current.type == NgSimpleTokenType.tagEnd ||
+        _current.type == NgSimpleTokenType.voidCloseTag) {
+      return scanElementEndOpen();
+    } else if (_current.type == NgSimpleTokenType.whitespace) {
```
MichaelRFairhurst, line 170: I think this is the downside of putting the error recovery in the lexer. Ideally the lexer wouldn't care about the order of tokens, it would just provide them. But I guess that's what the scanner does, though usually there isn't two ways of getting a token stream in a compiler.

Would it be possible to move this logic into the parser? then we wouldn't need this class at all, maybe providing a speed boost, and definitely making the architecture a bit more traditional.

lib/src/scanner.dart
```diff
@@ -2,326 +2,622 @@
...
+      _state = NgScannerState.scanAfterComment;
+      return new NgToken.commentValue(_current.offset, _current.lexeme);
     }
-    _state = _NgScannerState.scanAfterComment;
-    return new NgToken.commentValue(offset, _scanner.substring(offset));
+    return handleError();
+  }
+
+  // Doesn't switch states or check validity of current token.
+  NgToken _scanCompoundDecorator() {
```
MichaelRFairhurst, line 282: this strikes me also as something that belongs in the parser.

lib/src/scanner.dart
```diff
@@ -2,326 +2,622 @@
...
+    if (_current.type == NgSimpleTokenType.period ||
+        _current.type == NgSimpleTokenType.identifier) {
+      _state = NgScannerState.scanSuffixEvent;
+      return _scanCompoundDecorator();
+    }
+    return handleError();
+  }
+
+  @protected
+  NgToken scanSpecialPropertyDecorator() {
```
MichaelRFairhurst, line 473: This is an example of where moving this logic to the parser would clean up the code. The lexer has completely distinct states, and this logic is duplicated in each. But in the parser, it's just in "parseDecorator" and it can say "if this or that or otherwise, do x for all".

MichaelRFairhurst (approved):
Looks awesome. I have a comment somewhere in there where I went on a long schpiel (sic) over how it made me think about theory and how your solution is like solving an even bigger problem than just our parser.

lib/src/ast/element.dart
mk13: Yeah I agree. This is more of a temporary solution I wanted to implement as a placeholder before doing some more modifications on the AST level. Once I establish the AST to have better offset ranges, I'll remove the whitespaces list.

lib/src/ast/offset_info.dart
mk13: Sure, will change it to `TagOffsetInfo`.

lib/src/ast/property.dart
mk13: It shouldn't have to. `elementDecoratorToken` will never be null. If error recovery is disabled, the lack of identifier token will raise error immediately and exit out before even accessing this part of code. If error recovery is enabled, a synthetic is placed instead.

lib/src/ast/property.dart
mk13: "prefixOffset" much cleaner?

lib/src/ast/sugar/banana.dart
mk13: It's used during ElementAst comparison as well as against test-case synthetics. Since synthetics don't have any offset, it would throw test cases off.

lib/src/parser/reader.dart
mk13: This reader is meant to be used generically for both NgToken and NgSimpleToken - so null would be the best option still. However, I definitely want to revisit this two layer Token system and integrate them into a more uniform way - in which case I can return a single EOF token.

lib/src/exception_handler/exception_handler.dart
mk13: Let's discuss this in person - I want to pick your brain regarding this issue.

lib/src/parser/recursive.dart
mk13: Yes at the current moment. However, I think we can change this to not allow whitespaces in between - which would make the parsing a lot easier at this step too. @matanlurey  what do you think?

lib/src/recovery_protocol/angular_analyzer_protocol.dart
mk13: Yeah I think it would be better that way too. I'll probably switch it with the next reformatting/fixing.

lib/src/recovery_protocol/recovery_protocol.dart
mk13: Technically this part should never be hit in the first place; I just put this here to get rid of the pesky 'doesn't return with a ...' message.

lib/src/recovery_protocol/angular_analyzer_protocol.dart
mk13: Interesting. I didn't even think of it this way (largely due to lack of knowledge of how RegEx works internally). I think it would be cool to see if we can find a more systematic/shorthand way to implement this type of recovery.

mk13:
Fixes added. Still need to do: 

- Error storage and error recovery objects.
- Remove 'whitespaces' list in ElementAst.
- Figure out what can be moved from scanner into parser for better efficiency and flow.

These fixes will occur in a separate PR.

PR #26, opened 2017-03-07
Ast recovery

mk13:
Changes in this PR: 
- Error handling of dangling open/close elements that occurs at the AST level. Synthetics added as placeholders. Refer to `recover_errors_parser.dart` for sample cases.
- Added `CloseElementAst` which is linked to `ElementAst` as a field. This separates the open component and the close component - allowing for better insertion of synthetics and error recovery.
- Added CLI tester for ast level parsing, refer to `ast_cli_tester.dart`.
- Added randomized token generator and parser to check for edge cases as well as benchmarking. Could still be improved in future iteration. Refer to `test/random_generator_test/...`
- Error handling at the ExpressionAst.parse level. If error-recovery is enabled, ast will continue to be parsed even if expression parsing fails. 
- Related to above, `String value` is added as a field to any AST that holds an expressionAst field. This ensures that there is still a link to the String version of the value even if the ExpressionAst fails to parse (in which case it would be null). 

Needed changes:
- Better handle void element cases. Not all void elements are tracked by list in `parser.dart` file. 
- Improve on error recovery of NgContent and Templates(after desugaring occurs).

Future changes for better performance:
- Change Token types to enums rather than string.
- Change error accumulation to not use Exception class, but a custom lightweight class.

### DIFF
--- a/lib/src/ast/close_element.dart
+++ b/lib/src/ast/close_element.dart
@@ -1,0 +1,139 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:angular_ast/src/ast.dart';
+import 'package:angular_ast/src/token/tokens.dart';
+import 'package:angular_ast/src/visitor.dart';
+import 'package:source_span/source_span.dart';
+import 'package:quiver/core.dart';
+
+/// Represents the closing DOM element that was parsed.
+///
+/// Clients should not extend, implement, or mix-in this class.
+abstract class CloseElementAst implements StandaloneTemplateAst {
+  /// Creates a synthetic close element AST.
+  factory CloseElementAst(
+    String name, {
+    ElementAst openComplement,
+    List<WhitespaceAst> whitespaces,
+  }) = _SyntheticCloseElementAst;
+
+  /// Creates a synthetic close element AST from an existing AST node.
+  factory CloseElementAst.from(
+    TemplateAst origin,
+    String name, {
+    ElementAst openComplement,
+    List<WhitespaceAst> whitespaces,
+  }) = _SyntheticCloseElementAst.from;
+
+  /// Creates a new close element AST from a parsed source
+  factory CloseElementAst.parsed(
+    SourceFile sourceFile,
+    NgToken closeTagStart,
+    NgToken nameToken,
+    NgToken closeTagEnd, {
+    ElementAst openComplement,
+    List<WhitespaceAst> whitespaces,
+  }) = ParsedCloseElementAst;
+
+  @override
+  bool operator ==(Object o) {
+    if (o is CloseElementAst) {
+      return name == o.name && openComplement == o.openComplement;
+    }
+    return false;
+  }
+
+  @override
+  int get hashCode => hash2(name, openComplement);
+
+  @override
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
+    return visitor.visitCloseElement(this, context);
+  }
+
+  /// Whether this is a `</template>` tag and should be directly rendered.
+  bool get isEmbeddedTemplate => name == 'template';
+
+  /// Name (tag) of the close element.
+  String get name;
+
+  /// [ElementAst] that represents the open complement.
+  ElementAst get openComplement;
+
+  /// Whitespaces at the end
+  List<WhitespaceAst> get whitespaces;
+
+  @override
+  String toString() {
+    final buffer = new StringBuffer('$CloseElementAst <$name> { ');
+    if (openComplement != null) {
+      buffer..write('openComplement=')..write(openComplement)..write(' ');
+    }
+    if (whitespaces.isNotEmpty) {
+      buffer
+        ..write('whitespaces=')
+        ..writeAll(whitespaces, ', ')
+        ..write(' ');
+    }
+    return (buffer..write('}')).toString();
+  }
+}
+
+/// Represents a real, non-synthetic DOM close element that was parsed,
+/// that could be upgraded.abstract
+///
+/// Clients should not extned, implement, or mix-in this class.
+class ParsedCloseElementAst extends TemplateAst with CloseElementAst {
+  /// [NgToken] that represents the identifier tag in `</tag>`.
+  final NgToken identifierToken;
+
+  ParsedCloseElementAst(
+    SourceFile sourceFile,
+    NgToken closeElementStart,
+    this.identifierToken,
+    NgToken closeElementEnd, {
+    ElementAst openComplement,
+    this.whitespaces: const [],
+  })
+      : openComplement = openComplement,
+        super.parsed(closeElementStart, closeElementEnd, sourceFile);
+
+  @override
+  String get name => identifierToken.lexeme;
+
+  /// Complement open element ast
+  @override
+  ElementAst openComplement;
+
+  /// Whitespaces
+  @override
+  final List<WhitespaceAst> whitespaces;
+}
+
+class _SyntheticCloseElementAst extends SyntheticTemplateAst
+    with CloseElementAst {
+  _SyntheticCloseElementAst(
+    this.name, {
+    this.openComplement,
+    this.whitespaces: const [],
+  });
+
+  _SyntheticCloseElementAst.from(
+    TemplateAst origin,
+    this.name, {
+    this.openComplement,
+    this.whitespaces: const [],
+  })
+      : super.from(origin);
+
+  @override
+  final String name;
+
+  @override
+  ElementAst openComplement;
+
+  @override
+  final List<WhitespaceAst> whitespaces;
+}
--- a/lib/src/ast/close_element.dart
+++ b/lib/src/ast/close_element.dart
@@ -6,7 +6,6 @@ import 'package:angular_ast/src/ast.dart';
 import 'package:angular_ast/src/token/tokens.dart';
 import 'package:angular_ast/src/visitor.dart';
 import 'package:source_span/source_span.dart';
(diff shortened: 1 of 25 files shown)


lib/src/ast/close_element.dart
```diff
@@ -0,0 +1,119 @@
...
+  }) = _SyntheticCloseElementAst;
+
+  /// Creates a synthetic close element AST from an existing AST node.
+  factory CloseElementAst.from(
+    TemplateAst origin,
+    String name, {
+    List<WhitespaceAst> whitespaces,
+  }) = _SyntheticCloseElementAst.from;
+
+  /// Creates a new close element AST from a parsed source
```
liuming0: A dot after "source".

lib/src/ast/close_element.dart
```diff
@@ -0,0 +1,119 @@
...
+    }
+    return false;
+  }
+
+  @override
+  int get hashCode => name.hashCode;
+
+  @override
+  /*=R*/ accept/*<R, C>*/(TemplateAstVisitor/*<R, C>*/ visitor, [C context]) {
+    return visitor.visitCloseElement(this, context);
```
liuming0: How about use "=>"

lib/src/ast/close_element.dart
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /// Whether this is a `</template>` tag and should be directly rendered.
+  bool get isEmbeddedTemplate => name == 'template';
+
+  /// Name (tag) of the close element.
+  String get name;
+
+  /// Whitespaces at the end
```
liuming0: dot

lib/src/ast/close_element.dart
```diff
@@ -0,0 +1,119 @@
...
+  List<WhitespaceAst> get whitespaces;
+
+  @override
+  String toString() {
+    final buffer = new StringBuffer('$CloseElementAst </$name> { ');
```
liuming0: How about use "var buffer"

lib/src/ast/property.dart
```diff
@@ -66,6 +69,9 @@ abstract class PropertyAst implements TemplateAst {
   /// Name of the property being set.
   String get name;
 
+  /// Unquoted value being bound to property
```
liuming0: dot

lib/src/parser/recursive.dart
```diff
@@ -40,6 +41,30 @@ class RecursiveAstParser {
     return nodes;
   }
 
+  CloseElementAst parseCloseElement(NgToken beginToken) {
```
liuming0, line 44: Prefer use "var" for local variable.

test/ast_cli_tester.dart
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+import 'dart:convert';
+import 'package:angular_ast/angular_ast.dart';
+
+RecoveringExceptionHandler exceptionHandler = new RecoveringExceptionHandler();
+//ThrowingExceptionHandler exceptionHandler = new ThrowingExceptionHandler();
```
liuming0: Delete?

test/ast_cli_tester.dart
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+import 'dart:convert';
+import 'package:angular_ast/angular_ast.dart';
+
+RecoveringExceptionHandler exceptionHandler = new RecoveringExceptionHandler();
```
liuming0: Is exceptionHandler going to change? If not, "final exceptionHandler" should be used.

test/ast_cli_tester.dart
```diff
@@ -0,0 +1,50 @@
...
+//ThrowingExceptionHandler exceptionHandler = new ThrowingExceptionHandler();
+
+List<StandaloneTemplateAst> parse(String template) {
+  return const NgParser().parsePreserve(
```
liuming0: How about "=>"

test/ast_cli_tester.dart
```diff
@@ -0,0 +1,50 @@
...
+    template,
+    sourceUrl: '/test/parser_test.dart#inline',
+    exceptionHandler: exceptionHandler,
+  );
+}
+
+void main() {
```
liuming0: Quote should be consistent. Prefer single over double.
e.g. print('--------');

test/ast_cli_tester.dart
```diff
@@ -0,0 +1,50 @@
...
+    }
+    if (exceptionHandler is RecoveringExceptionHandler) {
+      final exceptionsList = exceptionHandler.exceptions;
+      if (exceptionsList.isEmpty) {
+        print("CORRECT!");
+      } else {
+        HumanizingTemplateAstVisitor visitor =
+            const HumanizingTemplateAstVisitor();
+        String fixed = ast.map((t) => t.accept(visitor)).join('');
+        print("ORGNL: " + input);
```
liuming0: print('ORGNL: $input');

liuming0 (approved):

lib/src/ast/close_element.dart
```diff
@@ -0,0 +1,118 @@
...
+      visitor.visitCloseElement(this, context);
+
+  /// Whether this is a `</template>` tag and should be directly rendered.
+  bool get isEmbeddedTemplate => name == 'template';
+
+  /// Name (tag) of the close element.
+  String get name;
+
+  /// Whitespaces at the end.
+  List<WhitespaceAst> get whitespaces;
```
MichaelRFairhurst: Once again -- do we need to track whitespaces? Is it possible to have multiple?

lib/src/ast/element.dart
```diff
@@ -45,11 +51,11 @@ abstract class ElementAst implements StandaloneTemplateAst {
...
   factory ElementAst.parsed(
     SourceFile sourceFile,
-    NgToken beginToken,
+    NgToken openElementStart,
     NgToken nameToken,
-    int openTagEnd,
-    int closeTagStart,
-    NgToken endToken, {
+    NgToken openElementEnd, {
+    bool isVoidElement,
```
MichaelRFairhurst: do we need to track both things? That its void and its close element? Couldn't the former be a getter which returns `closeComplement == null`?

lib/src/ast/element.dart
```diff
@@ -97,6 +109,20 @@ abstract class ElementAst implements StandaloneTemplateAst {
   /// Whether this is a `<template>` tag and should not be directly rendered.
   bool get isEmbeddedTemplate => name == 'template';
 
+  /// Determines whether the element tag name is void element.
+  bool get isVoidElement;
+
+  /// Whether this uses the void tag end '/>'.
+  ///
+  /// By definition, a void element can use either '>' or '/>'.
+  bool get usesVoidTagEnd;
```
MichaelRFairhurst: is it possible to get `/ >` and do we treat that the same? In other words, do we need an offset for this?

lib/src/ast/element.dart
```diff
@@ -207,12 +236,17 @@ class ParsedElementAst extends TemplateAst with ElementAst {
     this.stars: const [],
     this.whitespaces: const [],
   })
-      : super.parsed(beginToken, endToken, sourceFile);
+      : usesVoidTagEnd = openElementEnd.type == NgTokenType.openElementEndVoid,
```
MichaelRFairhurst: this could also be a getter rather than calculated & stored as a member

lib/src/ast/property.dart
```diff
@@ -127,13 +133,10 @@ class ParsedPropertyAst extends TemplateAst
     this.nameToken,
     this.suffixToken, [
     this.valueToken,
+    this.expression,
```
MichaelRFairhurst, line 136: +1 to this change.

I don't like that it used to fire a dart parser to create a node! I the expression being passed in

lib/src/parser/recursive.dart
```diff
@@ -103,23 +128,29 @@ class RecursiveAstParser {
...
           beginToken,
           prefixToken,
           decoratorToken,
           suffixToken,
           valueToken,
+          expressionAst,
           equalSignToken,
         );
       } else if (prefixType == NgTokenType.propertyPrefix) {
+        ExpressionAst expressionAst =
```
MichaelRFairhurst, line 146: again, +1 to this change! I like the organization a lot

MichaelRFairhurst (approved):
Nice change! Some nits but I love this change.

lib/src/ast/close_element.dart
mk13: Unfortunately this isn't the PR where I will be removing this - but will be coming soon shortly.

lib/src/ast/element.dart
mk13: Misread the question. I'll try this out to see if it's still viable.

lib/src/ast/element.dart
mk13: Yes that should be doable.

lib/src/ast/element.dart
mk13: Ugly 'usesVoidTagEnd' will be removed shortly.

lib/src/ast/close_element.dart
matanlurey: Maybe add a comment that this will be removed?

Or use `@Deprecated` already :)

matanlurey (approved):
Just tried it out, looks good :)

mk13:
Whitespaces now marked as to be deprecated. 'usesVoidTag' boolean value removed and all affecting functions also modified.

mk13:
'isVoidElement' is no longer a field variable and a getter based on closeComplement == null instead.

PR #27, opened 2017-03-09
Lightweight exceptions

mk13:
Created custom exception class to be used instead of standard Exception - goal is to reduce the weight of transferring exceptions. Previous exception was designed was for simple throwing - not accumulation. 

Future considerations:
- Implementing ErrorCode(Analyzer) and adding to every exception.

### DIFF
--- a/lib/src/exception_handler/exception_handler.dart
+++ b/lib/src/exception_handler/exception_handler.dart
@@ -2,15 +2,16 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:angular_ast/src/exception_handler/angular_parser_exception.dart';
 import 'package:meta/meta.dart';
 
 abstract class ExceptionHandler {
-  void handle(Exception e);
+  void handle(AngularParserException e);
 }
 
 class ThrowingExceptionHandler implements ExceptionHandler {
   @override
-  void handle(Exception e) {
+  void handle(AngularParserException e) {
     throw e;
   }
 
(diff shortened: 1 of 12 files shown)


MichaelRFairhurst (approved):
Seems like they should probably use an error type rather than strings, otherwise we won't be able to easily merge them into our own analysis errors.

They could also use offset+length rather than offset+lexeme, though that difference seems not super important to me

mk13:
I had similar thoughts :) This is just the precursor - expect to see another PR within the week with ErrorCodes added into it to make the error types more distinct and also more easily mergeable into angular_analyzer.

I'll also switch over to offset in that same iteration. After doing some analysis - you're correct, there really is no use in holding the actual lexeme in the exception since whatever uses these exceptions have access to the raw text and can simply get the string on demand. This should lead to more lightweight exceptions.

matanlurey:
Let me know when you rebase and I'll review.

mk13:
Hi @matanlurey, I've merged the previous PR#26 and this is now ready for review. Thanks!

lib/src/exception_handler/exception_handler.dart
```diff
@@ -20,10 +21,10 @@ class ThrowingExceptionHandler implements ExceptionHandler {
 }
 
 class RecoveringExceptionHandler implements ExceptionHandler {
-  final List<Exception> exceptions = new List<Exception>();
+  final List<AngularParserException> exceptions = <AngularParserException>[];
```
liuming0: List<AngularParserException> can be removed.

lib/src/scanner.dart
```diff
@@ -33,8 +34,11 @@ class NgScanner {
     return _current;
   }
 
-  factory NgScanner(String html, ExceptionHandler exceptionHandler,
-      {sourceUrl}) {
+  factory NgScanner(
+    String html,
+    ExceptionHandler exceptionHandler, {
+    sourceUrl,
```
liuming0: What is the type of sourceUrl?

matanlurey (approved):

PR #28, opened 2017-03-16
Mustache recovery & Ngcontent

mk13:
- Interpolation recovery logic added.
- 'ng-content' error recovery added.

### DIFF
--- a/lib/src/parser/recursive.dart
+++ b/lib/src/parser/recursive.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:collection';
+
 import 'package:angular_ast/src/ast.dart';
 import 'package:angular_ast/src/exception_handler/angular_parser_exception.dart';
 import 'package:angular_ast/src/parser/reader.dart';
@@ -42,6 +44,7 @@ class RecursiveAstParser {
     return nodes;
   }
 
+  /// Parses and returns a [CloseElementAst].
   CloseElementAst parseCloseElement(NgToken beginToken) {
     var nameToken = _reader.expect(NgTokenType.elementIdentifier);
     if (_voidElements.contains(nameToken.lexeme)) {
@@ -186,12 +189,13 @@ class RecursiveAstParser {
   /// Returns a DOM element AST starting at the provided token.
   ///
   /// It's possible the element will end up not being an [ElementAst].
-  StandaloneTemplateAst parseElement(NgToken beginToken) {
+  StandaloneTemplateAst parseElement(
+      NgToken beginToken, Queue<String> tagStack) {
     // Parse the element identifier.
     final nameToken = _reader.expect(NgTokenType.elementIdentifier);
     //TODO: do we want to allow trailing spaces in content/template tags?
     if (nameToken.lexeme == 'ng-content') {
-      return parseEmbeddedContent(beginToken);
+      return parseEmbeddedContent(beginToken, nameToken);
     } else if (nameToken.lexeme == 'template') {
       return parseEmbeddedTemplate(beginToken);
     }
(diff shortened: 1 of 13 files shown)


lib/src/parser/recursive.dart
```diff
@@ -253,39 +257,56 @@ class RecursiveAstParser {
...
+          var closeNameToken = _reader.peek();
+
+          if (closeNameToken.lexeme != nameToken.lexeme) {
+            // Found a closing tag, but not matching current [ElementAst].
+            exceptionHandler.handle(new AngularParserException(
+              'Invalid closing tag: $closeNameToken (expected $nameToken)',
+              closeNameToken.lexeme,
+              closeNameToken.offset,
+            ));
+            if (tagStack.contains(closeNameToken.lexeme)) {
```
MichaelRFairhurst, line 284: do you think this might fire too often for things like `<div>`?

So we might get like `<div><ul><li><span></div></li></ul></div>` and in this case it will be corrected into `<div><ul><li><span></span></li></ul></div><li></li><ul></ul><div></div>`.

On the other hand, I'm sure that absolutely every solution we come up with will have some holes.

Any ideas on how to simply solve the above case? Or do you think we should postpone for user testing?

lib/src/parser/recursive.dart
```diff
@@ -253,39 +257,56 @@ class RecursiveAstParser {
...
+              // If the closing tag is in the seen [ElementAst] stack,
+              // leave it alone. Instead create a synthetic close.
+              _reader.putBack(nextToken);
+              closeElementAst = new CloseElementAst(nameToken.lexeme);
+              closingTagFound = true;
+            } else {
+              // If the closing tag is not in the stack, create a synthetic
```
MichaelRFairhurst, line 291: this is a beautiful way to handle this

lib/src/parser/recursive.dart
```diff
@@ -307,41 +328,139 @@ class RecursiveAstParser {
...
+      equalSign,
+      valueToken,
     );
   }
 
+  /// Helper function that accumulates all parts of attribute-value variant
+  /// and returns it as a single string. Should be used to gather any
+  /// non-'select' decorator. Consumes all necessary erroneous tokens.
+  String _accumulateInvalidNgContentDecoratorValue(
+      NgToken beforeElementDecorator, NgToken nextToken) {
```
MichaelRFairhurst, line 437: Could this be approached more like "while token != end tag"?

lib/src/recovery_protocol/angular_analyzer_protocol.dart
```diff
@@ -218,7 +231,22 @@ class NgAnalyzerRecoveryProtocol extends RecoveryProtocol {
   @override
   RecoverySolution scanInterpolation(
       NgSimpleToken current, NgTokenReversibleReader reader) {
-    return new RecoverySolution.skip();
+    NgScannerState returnState;
+    NgToken returnToken;
+
+    if (current.type == NgSimpleTokenType.EOF ||
+        current.type == NgSimpleTokenType.commentBegin ||
+        current.type == NgSimpleTokenType.openTagStart ||
```
MichaelRFairhurst, line 239: I think in the future this will be valid dart:

``{{<int>["foo"].length}}``

lib/src/scanner.dart
```diff
@@ -237,7 +239,8 @@ class NgScanner {
       _state = NgScannerState.scanStart;
       return new NgToken.interpolationEnd(_current.offset);
     }
-    return handleError();
+    var overrideMessage = 'Unclosed mustache';
+    return handleError(overrideMessage: overrideMessage);
```
MichaelRFairhurst, line 243: could be just `overrideMessage: 'UnclosedMustache'`

lib/src/scanner.dart
```diff
@@ -270,7 +273,15 @@ class NgScanner {
...
       return new NgToken.interpolationStart(_current.offset);
     }
-    return handleError();
+
+    var override = _current;
+    if (_current.type == NgSimpleTokenType.text &&
+        _reader.peekType() == NgSimpleTokenType.mustacheEnd) {
+      override = _reader.peek();
+    }
+    var overrideMessage = 'Unopened mustache';
```
MichaelRFairhurst, line 282: Could use a comment on why this error case occurs. For instance, when would _current.type be text?

lib/src/simple_tokenizer.dart
```diff
@@ -206,7 +207,31 @@ class NgSimpleScanner {
...
+          _scanner.position = offset + mustacheMatch.end;
+          if (mustacheMatch.group(1) != null) {
+            return new NgSimpleToken.mustacheBegin(mustacheStart);
+          }
+          if (mustacheMatch.group(2) != null) {
+            return new NgSimpleToken.mustacheEnd(mustacheStart);
+          }
+        }
+        // Mustache doesn't exist; simple text.
+        return new NgSimpleToken.text(offset, text);
```
MichaelRFairhurst, line 234: very nice code, like the placement and implementation

test/recover_errors_lexer_test.dart
```diff
@@ -110,6 +112,39 @@ void main() {
...
+      new NgToken.interpolationStart(0), // Synthetic
+      new NgToken.interpolationValue(0, 'mustache text'),
+      new NgToken.interpolationEnd(13),
+    ]);
+    expect(recoveringException.exceptions.length, 1);
+    var e = recoveringException.exceptions[0];
+    expect(e.context, '}}');
+    expect(e.offset, 13);
+
+    expect(untokenize(results), '{{mustache text}}');
```
MichaelRFairhurst, line 144: Hmm, do we want to do this or do `mustache text{{}}`?

MichaelRFairhurst (approved):
LGTM, plus some minor suggestions/thoughts

lib/src/parser/recursive.dart
mk13: What about in the case of <ng-content blah="value" select="something" [prop]="value3">, this would mean that the the entire attribute name-value chains would be removed. Ideally I would want this to still keep the 'select="something"'.

lib/src/recovery_protocol/angular_analyzer_protocol.dart
mk13: Barriers for mustache error recovery is going to be fixed in a future PR (already done, just haven't submitted yet) to stop at mustache open, mustache close, EOF, and newline if the previous three cannot be found. It should be fine once that PR is submitted.

lib/src/scanner.dart
mk13: Good catch! I think for this scenario, I'm going to submit another PR after this that gets rid of string passing and instead use error codes. This should handle the below issue as well :)

lib/src/parser/recursive.dart
mk13: It would be cool to add a 'threshold' to see when to hold off in terms of stack length. For example, if it exceeds 2 stacks, it treats it as a dangling - in which case it would be resolved as:

`<div><ul><li><span></span><div></div></li></ul></div>` but this might have issues within it as well. As you mentioned, adding a threshold would prevent this from triggering too often and completely mangling the original.

I think user testing would be the best bet in finding the right balance between recovering too often and not recovering enough.

test/recover_errors_lexer_test.dart
mk13: This is a good question. We can try and recover using the preceding text, which is something what current angular_analyzer does, or treat it explicitly as a dangling and insert an open just before it. 

I think the former is a safer 'catch all'. In the case that the preceding text is not a valid expression (due to mistyped expression or meant to be text), it will continue placing the 'mustache text' chunk as both a text as well as a potential mustache. It would give us more flexibility by allowing mustache. 

On the other hand, the latter would only try and autocomplete expression if the cursor is right before the dangling close.

PR #30, opened 2017-03-29
Major changes - error codes, template recovery, decoupling of expression and desugaring

mk13:
Sorry for another major pull request - but this should be the last major one prior to full official integration into angular_analyzer_plugin.

**Major changes:** 
- Error recovery in EmbeddedTemplateAst.
- Desugaring visitor improved to also impact pipes.
- Expression parsing completely decoupled from desugaring; set as a separate visitor.
- Main 'parse' function modified accordingly to top two.
- Error codes introduced. Extends off of analyzer error code allowing for better integration into angular_analyzer.
- Dart expressions maintain offset by using padded whitespace and buffer added and expression parsing doesn't crash on padded whitespaces.
- Mustache error recovery improved.
- `<!DOCTYPE` ...> parsing handled.

**Bug fixes:** 
- EmbeddedContentAst bugs in parsing
- Keeping track of postfix and unit in event and property
- Empty comment no longer crashes
- Dangling quote bug fix revealed by angular_analyzer fuzz test
- Handling null exception moved to RecoveringExceptionHandler
- Bug where desugaring wouldn't work properly at child level


lib/src/ast/event.dart
```diff
@@ -159,7 +158,7 @@ class ParsedEventAst extends TemplateAst
   @override
   String get postfix {
     final split = _nameWithoutParentheses.split('.');
-    return split.length > 1 ? split.last : null;
+    return split.length > 1 ? split[1] : null;
```
MichaelRFairhurst: Hmm, doesn't this only work differently when there are multiple '.'? Is that presented as an error anywhere?

lib/src/exception_handler/exceptions.dart
```diff
@@ -0,0 +1,245 @@
...
+part of angular_ast.src.exceptions;
+
+/// Error codes used for exceptions that occur during the parsing.
+/// The convention for this class is for the name of the error code
+/// to indicate the problem that caused the error code to be generated
+/// and for the error message to explain what is wrong, and when appropriate,
+/// how the problem can be corrected.
+class NgParserWarningCode extends ErrorCode {
+  static const NgParserWarningCode AFTER_COMMENT = const NgParserWarningCode(
+    'AFTER_COMMENT',
```
MichaelRFairhurst: some of these names are a bit strange

lib/src/expression/micro.dart
```diff
@@ -21,12 +21,15 @@ bool isMicroExpression(String expression) {
...
+  String expression,
+  int expressionOffset, {
   String sourceUrl,
 }) {
+  var paddedExpression = ' ' * expressionOffset + expression;
   return const NgMicroParser().parse(
     directive,
-    expression,
+    paddedExpression,
+    expressionOffset,
```
MichaelRFairhurst: Hmm, if `NgMicroParser()` already needs `expressionOffset`, it probably makes more sense that it does the padding itself, so its less likely to be forgotten at other places its used

lib/src/expression/micro/scanner.dart
```diff
@@ -17,14 +17,21 @@ class NgMicroScanner {
...
 
   _NgMicroScannerState _state = _NgMicroScannerState.scanInitial;
 
   factory NgMicroScanner(String html, {sourceUrl}) {
     return new NgMicroScanner._(new StringScanner(html, sourceUrl: sourceUrl));
   }
 
-  NgMicroScanner._(this._scanner);
+  NgMicroScanner._(this._scanner) {
+    // Use padded whitespace to find real offset.
```
MichaelRFairhurst: Seems kinda hacky

lib/src/scanner.dart
```diff
@@ -160,18 +171,28 @@ class NgScanner {
...
+    // Only triggered by EOF.
+    return handleError(
+      NgParserWarningCode.AFTER_COMMENT,
+      _lastCommentStartOffset,
+      _current.offset - _lastCommentStartOffset,
+    );
   }
 
+  // TODO: Max: Better handle cases like 'prop]'. Instead of
+  // TODO: 'prop []' resolve, resolve as '[prop].
```
MichaelRFairhurst, line 183: Maybe that's something we can do in the parser? Like, on '=', we say, if synthetic, peek up to the first non-synthetic node, and if its ']', ')', or ')]', flip the lexeme and continue? Otherwise, putback all? Or maybe we can do a getNoErrorRecovery method/setState(afterCloseBracket)

lib/src/ast/event.dart
MichaelRFairhurst: I see the error down below.

Might be good to comment here...maybe the best would be some kind of assert. `assert(split.length < 2)` though that would mean where you report the error you'd have to trim the postfixes, which seems wrong. So...maybe best is to throw an exception when length is long, like, "no you figure out what you want to do with this garbage"?

lib/src/scanner.dart
```diff
@@ -224,30 +348,84 @@ class NgScanner {
...
+        type == NgSimpleTokenType.closeTagStart ||
+        type == NgSimpleTokenType.EOF) {
+      return handleError(
+        NgParserWarningCode.AFTER_ELEMENT_IDENTIFIER,
+        _lastToken.offset,
+        _lastToken.length,
+      );
+    }
+
+    return handleError(
```
MichaelRFairhurst, line 385: Hmm, this seems like pretty delicate code. For instance, if we add a new token type, we have to revisit all these spots.

What about putting these into a table? State -> TokenType -> Error. Then it'd be all in one place?

lib/src/visitors/expression_parser_visitor.dart
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/error/error.dart';
+import 'package:angular_ast/src/ast.dart';
+import 'package:angular_ast/src/exception_handler/exception_handler.dart';
+import 'package:angular_ast/src/visitor.dart';
+
+class ExpressionParserVisitor<C> implements TemplateAstVisitor<TemplateAst, C> {
```
MichaelRFairhurst: this can be TemplateAstVisitor<Null, C> and then your methods don't have to return (since the visitor does state modifications only)

MichaelRFairhurst (approved):
Looks great. Can really see it coming along, at this point its like just a bunch of edge cases.

Some of the constructs here have me a little worried about maintainability, though, like a lot of hand-crafted offsets in tests and carefully constructed in the parser and/or a visitor and/or the scanner etc. And definitely some new repetition.

But how do you feel about that? Do you feel like that's stuff that's done, was straightforward, shouldn't change much? Or do you feel like it'd be worth trying to come up with some models that simplify how many details we have to manage when we track it?

For instance, maybe merging the recovery solution with the error reporter, or having chains of them where the first one reports errors, or maybe having the parser report when it sees synthetics so that everything is in the parser.

Looks like you've put in a ton of hard work to iron out all the corners, so that is fantastic, and I only want to revisit stuff that might require more routine work. But I imagine this was an incredibly tough PR to finalize, so if there's more to be done then maybe we can find a way to make it easier to tweak next time.

lib/src/ast/event.dart
mk13: Sounds like a good plan. I'll likely throw the error at the scanner level since the event module has no knowledge of the exceptionHandler.

lib/src/exception_handler/exceptions.dart
mk13: Yeah, I initially set the name of the errors as the state of where they're thrown. They'll have to be refactored into more meaningful names. I also want to relook at the exception overlaps between angular_ast and angular_analyzer_plugin in a future rework of exception codes. 

I'll rename some of the exceptions for the moment.

lib/src/expression/micro/scanner.dart
mk13: Hmm unfortunately this seems the most efficient way since the _scanner utilizes the offsets directly. As a result, the _scanner has to be the one that eliminates the whitespaces and record the offset of where the true expression begins.

lib/src/scanner.dart
mk13: Sure, we can move this to the parser. The only concern I have with this is that it adds a lot of extra overhead into the primary converted NodeInfos we use in angular_analyzer_plugin (current implementation takes the overall AST and uses it at face value; doesn't dissect the individual components for synthetic parts).

lib/src/scanner.dart
mk13: Agreed. I've actually thought of using a table too that maps out errors as well as correct/standard path. I'll introduce that into a separate PR.

lib/src/visitors/expression_parser_visitor.dart
mk13: Sure, will change.

lib/src/ast/event.dart
mk13: Oh wait I'm already throwing it. It's in Recursive. Look for 'PROPERTY_NAME_TOO_MANY_FIXES' and 'EVENT_NAME_TOO_MANY_FIXES'.

matanlurey (approved):

lib/src/scanner.dart
MichaelRFairhurst: Yeah, I mean, its in a TODO anyway, mostly was just thinking out..loud, or...thinking...out...comment...

Issue #3, opened 2016-11-17
Support bind-*, on-* syntax for attributes, events


jonahwilliams:

mk13:
Working on this now.

mk13:
Resolved.

PR #33, opened 2017-04-14
On- bind- parsing and events in Templates

mk13:
Allow (event) bindings to exist in <template ...>.

'on-X' is parsed as an EventAst and 'bind-X' is parsed as a PropertyAst. (ref #3 )

Also added 'ParsedDecoratorAst' generic interface to be used by all Parsed decorator Asts. Moved all interfaces to 'interface.dart' as well for simplicity.


lib/src/ast/event.dart
```diff
@@ -85,14 +85,20 @@ abstract class EventAst implements TemplateAst {
 ///
 /// Clients should not extend, implement, or mix-in this class.
 class ParsedEventAst extends TemplateAst
-    with EventAst, TagOffsetInfo, SpecialOffsetInfo {
-  /// Token representing the `(property)` element decorator.
+    with EventAst, ParsedDecoratorAst, TagOffsetInfo {
```
MichaelRFairhurst: I think this would be cleaner with `implements` than using a mix in, since the mixed in classes don't define any behavior

lib/src/parser/recursive.dart
```diff
@@ -187,6 +187,41 @@ class RecursiveAstParser {
         );
       }
     }
+
+    // At this point, it is a TextAttribute, but handle cases
+    // with 'on-' and 'bind-' prefix.
+    if (decoratorToken.lexeme.startsWith('on-')) {
```
MichaelRFairhurst, line 193: Hmm, maybe `on-` and `bind-` could be tokens. Would that improve or harm the code flow do you think?

MichaelRFairhurst: Oh, and looks to me like this doesn't report an error for `on-="foo"`

MichaelRFairhurst (approved):
LGTM

lib/src/parser/recursive.dart
mk13: Adding 'on-' and 'bind-' to NgSimpleToken just adds unnecessary complexity without any functionality. (They do exist as an NgToken though.) For all intents and purposes, the syntactical functionality of 'on-' or 'bind-' is no more different than a standard text attribute parsing. As a result, I've decided to add this logic at the very last step of decorator AST conversion simply deriving off of text attribute.

I can add an error in this logic for if the term after on-/bind- is an empty string. Since there is no recovery needed on this error (An empty string is already there), it should be perfectly fine to keep this in the final step.

lib/src/ast/event.dart
mk13: Sure, will change.

lib/src/parser/recursive.dart
matanlurey: Since these are soft-deprecated anyway I agree with @mk13

matanlurey (commented):
Nice job, thanks!

PR #34, opened 2017-04-18
Condense element types - remove EmbeddedTemplate and EmbeddedContent

mk13:
This is a pretty big design change - I've decided to essentially remove 'EmbeddedTemplateAst' and 'EmbeddedContentAst'. 

TL;DR: Having three types had no purpose and just overcomplicated things.

Reasoning: 
- At a class level, these two weren't doing anything unique. They were both just essentially a 'gutted' version of 'ElementAst' with different class names. They're all just containers to begin with.
- Instead of using the class name to differentiate, just use a simple bool flag within the ElementAst (isTemplate and isNgContent).
- Building from above, this allows for a 'common type' where users won't have to always check if the type is ElementAst, EmbeddedContentAst, or EmbeddedTemplateAst. All the appropriate irrelevant fields are already emptied out (Example: if name is 'ng-content', bananas, events, properties, etc. are all empty lists).


ferhatb:
Have you looked at AngularDart source to see if change in visitors will be
ok?

A lot of this mirrors existing code to make migrating maybe easier?

mk13:
I'll take a look at the AngularDart code to see how they distinguish these types and see how their visitors behave.

MichaelRFairhurst:
Its also possible to create a base/mixin class which distinguishes these, ie,

```
abstract class ElementTypeDistinguishingVisitorMixin implements AngularAstVisitor {
  void visitElementAst(ElementAst ast) {
    if (ast.name == 'template') {
      visitEmbeddedTemplateAst(ast);
    } else if (ast.name == 'ng-content') {
      visitEmbeddedContentAst(ast);
    } else {
      visitNormalElementAst(ast);
    }
  }

  void visitEmbeddedTemplateAst(ElementAst ast);
  void visitEmbeddedContentAst(ElementAst ast);
  void visitNormalElementAst(ElementAst ast);
}

class PreexistingAngularVisitor with ElementTypeDistinguishingVisitorMixin {
   // hopefully relatively few changes here
}
```

Alternatively, all three classes could share an interface and we could use

```
abstract class HandleEmbeddedContentLikeNormalVisitorMixin implements TemplateAstVisitor {
    void visitCommonElement(CommonElementAst ast);

    void visitElementAst(ElementAst ast) {
      visitCommonElement(ast);
    }

    void visitEmbeddedContentAst(EmbeddedContentAst ast) {
      visitCommonElement(ast);
    }

    void visitEmbeddedTemplateAst(EmbeddedTemplateAst ast) {
      visitCommonElement(ast);
    }
}
```

and alternatively alternatively, we can use composition over inheritance if we want to prevent accidental overriding of the special mixin methods. (the "normalizing" visitor is a regular visitor, but accepts a "normalized" visitor which has no method `visitElementAst`, only a method `visitCommonElement`, and the normalizing visitor just forwards the visitations)

mk13:
I'm liking the mixin format, but feel that it might be overkill at the current implementation status since none of the visitors do anything special with Template or NgContent as of yet (maybe there might be visitors in the future that do so?). 

Expression parser visitor only looks at events and properties within Elements. (Interpolation and Text are separate ast-nodes). Desugar visitor doesn't do anything if the element isTemplate or isNgContent - although this may have to be changed depending on whether banana syntax is allowed in existing templates.

As per AngularDart, it seems they're also condensing all element types as a generic 'HtmlElementAst', and use a 'preparsedElement' on the actual AST and uses that to find PreparsedElementType.NG_CONTENT.

mk13:
I looked through the visitors in AngularDart and it seems that the only visitor that does things differently for the three element types is only in ViewBuilderVisitor. 

The others (ViewBinderVisitor, TemplateHumanizer, TemplateContentProjectionHumanizer) have a lot of overlapping behavior within all three types (ElementAst, NgContent, EmbeddedTemplateAst) and can realistically be merged into a single visitor (visitElementAst) that still maintains the intended behavior. 

If there are no plans to expand the number of visitors, then the Mixin might even be overkill since it will only realistically be applied to ViewBuilderVisitor.

ferhatb:
SG. If thats the only visitor, please go ahead with unifying.

lib/src/parser/recursive.dart
```diff
@@ -401,10 +363,56 @@ class RecursiveAstParser {
       tagStack.removeFirst();
     }
 
-    if (isTemplateElement) {
-      return new EmbeddedTemplateAst.parsed(
+    if (isNgContent) {
```
MichaelRFairhurst, line 366: What about using a subtyping pattern here?

```
class NgContentElement extends Element {

  void reportChildNode(Element childNode, ExceptionHandler handler) {
    // NgContentElement.reportChildNode throws while Element and TemplateElement accept it
  }
  ...
}
```

would remove some of the else-ifiness from the code. Although I suppose it would once again that the *instantiation* of these would have differences....Hmm, for that, maybe you could make a `factory` for `Element`.

```
factory Element(String tagname, ...) {
   if (tagname == "template") {
     return new TemplateElement(tagname, ...);
   } else ...
```

lib/src/parser/recursive.dart
```diff
@@ -425,136 +443,23 @@ class RecursiveAstParser {
         properties: properties,
         references: references,
         bananas: bananas,
-        stars: stars,
+        stars: (uniqueStarAst == null ? [] : [uniqueStarAst]),
```
MichaelRFairhurst, line 446: we may want to have all the star asts inside the ElementAst, so that we can, for instance, autocomplete inside them before forcing the user to refactor into two elements with stars.

MichaelRFairhurst (approved):
Apologies for blanking on this for so long. Seems I missed the moment when the original question was answered.

Can't argue its a nice reduction of code!

Did have some structural thoughts, but I think I might leave such structural decisions overall to you and @matanlurey . I can see some value in keeping things a bit more distinct...but definitely to me this looks like a quite nice maintanability change.

lib/src/parser/recursive.dart
mk13: I've tried the sub-typing pattern before but it still causes the issue of Element potentially being also either TemplateElement or NgContentElement - and we would just have to check those again. 

I haven't tried the factory solution yet, but will try once this issue is picked up again.

lib/src/parser/recursive.dart
mk13: Sounds like a good idea. I'll probably incorporate this in a separate PR with all the necessary checks needed for this.

mk13:
Putting this PR onto a freeze until more PoC can be done on how well it can integrate into angular2.

matanlurey (approved):

kevmoo:
Guessing we can close this down, right?

PR #35, opened 2017-04-18
List error codes

mk13:
Creates a list of error codes - to be used by angular-analyzer-plugin.


MichaelRFairhurst:
LGTM

Can you also (in this PR or another) make a function to look up the error code objects by error code name? In the dart analyzer they made a hashmap of codes to objects, and on lookup it lazily fills the map based on the list

MichaelRFairhurst (approved):

mk13:
I added that part to the angular_analyzer_plugin. Do you think we should condense the logic into angular_analyzer_plugin or keep it separate?
 
Refer to https://github.com/dart-lang/angular_analyzer_plugin/pull/276 for the part that improves the lazy loading.

PR #42, opened 2017-05-10
Mustache in plain attribute and improve flow of html-text mustache.

mk13:
Enabled mustache-parsing in attribute values with error-recovery. Similar to the parsing logic used in HTML-text level interpolation, but does not have EOF/EOL logic since it is encapsulated within quotes.

### DIFF
--- a/lib/src/ast/attribute.dart
+++ b/lib/src/ast/attribute.dart
@@ -66,7 +66,7 @@ abstract class AttributeAst implements TemplateAst {
 
   /// Mustaches found within value; may be `null` if value is null.
   /// If value exists but has no mustaches, will be empty list.
-  List<InterpolationAst> mustaches;
+  List<InterpolationAst> get mustaches;
 
   /// Static attribute value with quotes attached;
   /// may be `null` to have no value.
--- a/lib/src/ast/attribute.dart
+++ b/lib/src/ast/attribute.dart
@@ -2,13 +2,14 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:collection/collection.dart';
 import 'package:angular_ast/src/ast.dart';
 import 'package:angular_ast/src/token/tokens.dart';
 import 'package:angular_ast/src/visitor.dart';
 import 'package:source_span/source_span.dart';
 import 'package:quiver/core.dart';
 
-// TODO: Interpolation within the value
+final _listEquals = const ListEquality<dynamic>();
 
 /// Represents a static attribute assignment (i.e. not bound to an expression).
 ///
@@ -18,6 +19,7 @@ abstract class AttributeAst implements TemplateAst {
   factory AttributeAst(
     String name, [
     String value,
+    List<InterpolationAst> mustaches,
   ]) = _SyntheticAttributeAst;
 
   /// Create a new synthetic [AttributeAst] that originated from node [origin].
@@ -25,6 +27,7 @@ abstract class AttributeAst implements TemplateAst {
     TemplateAst origin,
     String name, [
     String value,
+    List<InterpolationAst> mustaches,
   ]) = _SyntheticAttributeAst.from;
 
   /// Create a new [AttributeAst] parsed from tokens from [sourceFile].
@@ -34,6 +37,7 @@ abstract class AttributeAst implements TemplateAst {
     NgToken nameToken, [
     NgAttributeValueToken valueToken,
     NgToken equalSignToken,
+    List<ParsedInterpolationAst> mustaches,
   ]) = ParsedAttributeAst;
 
   @override
@@ -44,7 +48,9 @@ abstract class AttributeAst implements TemplateAst {
   @override
   bool operator ==(Object o) {
     if (o is AttributeAst) {
-      return name == o.name && value == o.value;
+      return name == o.name &&
+          value == o.value &&
+          _listEquals.equals(mustaches, o.mustaches);
     }
     return false;
   }
@@ -57,6 +63,10 @@ abstract class AttributeAst implements TemplateAst {
 
   /// Static attribute value; may be `null` to have no value.
   String get value;
+
+  /// Mustaches found within value; may be `null` if value is null.
+  /// If value exists but has no mustaches, will be empty list.
+  List<InterpolationAst> mustaches;
 
   /// Static attribute value with quotes attached;
   /// may be `null` to have no value.
@@ -96,6 +106,7 @@ class ParsedAttributeAst extends TemplateAst
     this.nameToken, [
     this.valueToken,
     this.equalSignToken,
+    this.mustaches,
   ])
       : super.parsed(
           beginToken,
@@ -118,6 +129,10 @@ class ParsedAttributeAst extends TemplateAst
   /// Static attribute value; may be `null` to have no value.
   @override
   String get value => valueToken?.innerValue?.lexeme;
+
+  /// Parsed static attribute parts that are mustache-expressions.
+  @override
+  final List<InterpolationAst> mustaches;
 
   /// Static attribute value including quotes; may be `null` to have no value.
   @override
@@ -153,14 +168,18 @@ class _SyntheticAttributeAst extends SyntheticTemplateAst with AttributeAst {
   final String value;
 
   @override
+  final List<InterpolationAst> mustaches;
+
+  @override
   String get quotedValue => value == null ? null : '"$value"';
 
-  _SyntheticAttributeAst(this.name, [this.value]);
+  _SyntheticAttributeAst(this.name, [this.value, this.mustaches]);
 
   _SyntheticAttributeAst.from(
     TemplateAst origin,
     this.name, [
     this.value,
+    this.mustaches,
   ])
       : super.from(origin);
 }
--- a/lib/src/ast/attribute.dart
+++ b/lib/src/ast/attribute.dart
@@ -37,7 +37,7 @@ abstract class AttributeAst implements TemplateAst {
     NgToken nameToken, [
     NgAttributeValueToken valueToken,
     NgToken equalSignToken,
-    List<ParsedInterpolationAst> mustaches,
+    List<InterpolationAst> mustaches,
   ]) = ParsedAttributeAst;
 
   @override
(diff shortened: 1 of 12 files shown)


lib/src/ast/attribute.dart
```diff
@@ -2,13 +2,14 @@
...
 
+import 'package:collection/collection.dart';
 import 'package:angular_ast/src/ast.dart';
 import 'package:angular_ast/src/token/tokens.dart';
 import 'package:angular_ast/src/visitor.dart';
 import 'package:source_span/source_span.dart';
 import 'package:quiver/core.dart';
 
-// TODO: Interpolation within the value
+final _listEquals = const ListEquality<dynamic>();
```
liuming0: new ListEquality<dynamic>()

MichaelRFairhurst: I'd recommend making this a `ListEquality<Mustache>` and calling it like `_mustacheListEquals`

lib/src/ast/element.dart
```diff
@@ -9,7 +9,7 @@ import 'package:collection/collection.dart';
 import 'package:source_span/source_span.dart';
 import 'package:quiver/core.dart';
 
-const _listEquals = const ListEquality();
+const _listEquals = const ListEquality<dynamic>();
```
MichaelRFairhurst, line 12: is this needed?

test/parser_test.dart
```diff
@@ -503,6 +515,43 @@ void main() {
     expect(interpolation.expression, isNotNull);
   });
 
+  test(
+      'should parse, parse expressions, and preserve strict offsets for interpolations in values',
```
liuming0: This line is too long.

liuming0 (approved):

lib/src/parser/recursive.dart
```diff
@@ -531,6 +539,111 @@ class RecursiveAstParser {
...
+              NgParserWarningCode.UNOPENED_MUSTACHE,
+              mustacheEnd,
+              '}}'.length,
+            ));
+            mustaches.add(new InterpolationAst.parsed(
+              _source,
+              new NgToken.generateErrorSynthetic(
+                  mustacheBegin, NgTokenType.interpolationStart),
+              new NgToken.interpolationValue(
+                  mustacheBegin, text.substring(position, matchPosition)),
```
MichaelRFairhurst, line 617: I'm not sure about this. If I write

```
 blah blah this content was here
 iAmDeletingThisBindingStartingWithItsOpenBraces}}
```

then I'm not sure I'd want to see syntax errors for "blah blah blah" (and of course, knowing to start at "iAmDeleting..." is nontrivial). I think this kind of fits with the whole "left-side-best-side" philosophy, I think we should use an empty value since we can't determine where the mustache was supposed to begin -- rather, assume that the left side not being in a mustache is correct, and flag the right side, the unopened mustache, only.

lib/src/parser/recursive.dart
```diff
@@ -531,6 +539,111 @@ class RecursiveAstParser {
...
+              new NgToken.interpolationEnd(mustacheEnd),
+            ));
+          }
+        }
+        scanner.position = position + match.end;
```
MichaelRFairhurst: not +=?

lib/src/visitors/expression_parser_visitor.dart
```diff
@@ -15,7 +15,13 @@ class ExpressionParserVisitor<C> implements TemplateAstVisitor<Null, C> {
       : exceptionHandler = exceptionHandler ?? new ThrowingExceptionHandler();
 
   @override
-  visitAttribute(AttributeAst astNode, [_]) => null;
+  visitAttribute(AttributeAst astNode, [_]) {
+    if (astNode.mustaches != null) {
+      astNode.mustaches.forEach((mustache) {
```
MichaelRFairhurst, line 20: I might prefer `=>` here

lib/src/visitors/expression_parser_visitor.dart
```diff
@@ -45,6 +51,9 @@ class ExpressionParserVisitor<C> implements TemplateAstVisitor<Null, C> {
 
   @override
   visitElement(ElementAst astNode, [_]) {
+    astNode.attributes.forEach((attribute) {
```
MichaelRFairhurst, line 55: also might prefer `=>`

test/parser_test.dart
```diff
@@ -503,6 +515,43 @@ void main() {
...
+
+    expect(element.attributes.length, 1);
+    var attr = element.attributes[0] as ParsedAttributeAst;
+    expect(attr, new isInstanceOf<ParsedAttributeAst>());
+    expect(attr.value, '{{ 1 + 2 }} nonmustache {{ 3 + 4 }}');
+
+    expect(attr.mustaches.length, 2);
+    var mustache1 = attr.mustaches[0] as ParsedInterpolationAst;
+    var mustache2 = attr.mustaches[1] as ParsedInterpolationAst;
+    expect(mustache1.beginToken.offset, 15);
```
MichaelRFairhurst: I would do `templateString.indexOf("{{ 1 + 2 }}")`

test/parser_test.dart
```diff
@@ -503,6 +515,43 @@ void main() {
...
+    expect(mustache1.beginToken.lexeme, '{{');
+    expect(mustache1.beginToken.errorSynthetic, false);
+    expect(mustache1.valueToken.offset, 17);
```
MichaelRFairhurst: for all these offsets

MichaelRFairhurst (approved):
biggest possible changes I see are the offsets in the tests being fragile, and the recovery strategy being maybe a bit ambitious. But it is nicely readable and I saw no bugs in it, nice tests too.

lib/src/ast/element.dart
mk13: Not necessarily, but the IDE was complaining about it not having a type defined. Wanted the silence the IDE.

lib/src/parser/recursive.dart
mk13: Just making sure I got this correct:
` blah blah blah this is content 
thisWasSupposedToBeWrapped}}`

should be interpreted as: 
` blah blah blah this is content 
thisWasSupposedToBeWrapped{{}}`
while raising error on the '}}'. I'm guessing at the same time there wouldn't be any point in having an empty-string InterpolationAst at this point too.

lib/src/parser/recursive.dart
mk13: Doh.

lib/src/visitors/expression_parser_visitor.dart
mk13: Won't be able to change these (as discussed in person) because 1.21 on travis likes to complain when using this.

lib/src/parser/recursive.dart
mk13: After mulling this over, I think your suggested approach with a slight tweak is the best way to resolve dangling mustaches. One of the largest pitfalls of mustaches in particular is that we're dealing with dart expressions and potentially recovering too hard can lead to many pitfalls. 

The alternative - not recovering but simply marking danglings - is a far safer option since the intended expressions will just be treated as plain text... which has no real consequences aside from the intended behavior not occurring (which is fine considering there is an error). 

However, the one exception to this will be if the enclosing space of a mustache is only whitespace, I'll treat that space as enclosing mustache. 

For example: 
```
<someTag someAttr="blah blah {{    ">
```
In this scenario, the space following the '{{' will be treated as '{{    ' since it's fairly safe to treat this space as mustache. It will provide significant benefit in terms of autocompletion since a user might have his/her text cursor in that space and may be in the process of adding content.

On the other hand: 

```
<someTag someAttr="blah blah {{ blah blah">
```

In this scenario, ' blah blah' follows the dangling mustache - which may or may not be intended to be part of the mustache. Assuming it is NOT is a safer option so the overall attribute value will be treated as plain text with the error thrown in indicating the dangling mustache.

This will be similar for dangling '}}'. 

A more complicated example:

```
<someTag someAttr=" }} blah {{     {{ blah">
```

In this scenario, there's only two recovered mustaches: 
' }}' and '{{     '(first dangling open). 
'{{ blah' will NOT be recovered.


HTML-text interpolation will have the one extra step of accommodating for newline:

```
<div>
    Age: {{
    Name: Bob
</div>

```

In this scenario, there will still be an interpolation at the end of the '{{' since it is a dangling mustache that encounters a newline. This means that a user can autocomplete if the cursor follows after the '{{' while on the same line. 

On the other hand:

```
<div>
    Age: {{ something here
    Name: Bob
</div>
```
In this scenario, the '{{ something here' will be treated as plain-text instead of a mustache expression until the user manually closes it.

MichaelRFairhurst: Interesting.

We had this same problem in our code, and I remember that our first approach (assuming everything after {{ is intentionally a binding so we could autocomplete) caused problems when there was a space. The issue was that `{{ Bob` was tokenized and parsed, but the tokenizer drops whitespace. So it was treated as `{{ [DART_HERE]Bob[/DART_HERE]`. This meant that the `{{` was completely detached from the recognized dart code, and wasn't autocompletable. We changed it to only parse when its not followed by a space ie `{{Bob`.

The dart team said what we should do is make a synthetic parenthesized expression so that the dart code can contain the whitespace. If we do this, we don't need to worry about spaces. But if we don't, we do need to worry about those spaces preventing autocomplete.

mk13: That sounds reasonable. We can add synthetic parentheses around the whitespace without impacting autocompletion offset by doing a quick synthetic check on surrounding '{{' or '}}'. 

However, with the new AST we don't have to worry about whitespaces being dropped because InterpolationAst has two values we can use: 'value' and 'expression'. 'value' is the raw-text format within the brackets while 'expression' is the ExpressionAst object formed after parse is called on the text.

MichaelRFairhurst: the issue is that parsing `_    Bob` results in a dart Expression object that doesn't contain the space. So we can't suggest anything at that location. Only by parsing `( Bob)` could we suggest things before the space.

However, putting `(` around everything could lead to some very strange errors. For instance, `{{foo) + (1}}` wouldn't report an error at all

mk13: Let's have a face to face discussion to work out some of the issues :)

PR #46, opened 2017-05-22
Add let binding

mk13:
References aren't supposed to behave in the same way as 'let-' bindings in many ways. Instead of treating them as pseudo-references, making them more explicit and creating an AST will overall improve structure of AST.

Changed:
- Added new `LetBindingAst`. Can only exist within `EmbeddedTemplateAst`; if attempted to be used in non-embedded template, it will throw an error.
- Desugaring now creates `LetBindingAst` synthetics instead of `ReferenceAst` synthetic. This is important because `let-` binding and `#hashRef` behaves very differently when used in an explicit `<template>` vs star-syntax being desugared.
- Warns of empty `let-` bindings; ex: `<template let-></template>`


lib/src/exception_handler/exceptions.dart
```diff
@@ -162,6 +162,10 @@ class NgParserWarningCode extends ErrorCode {
       const NgParserWarningCode('INVALID_DECORATOR_IN_TEMPLATE',
           "Invalid decorator in 'template' element");
 
+  static const NgParserWarningCode INVALID_LET_BINDING_IN_NONTEMPLATE =
+      const NgParserWarningCode('INVALID_LET_BINDING_IN_NONTEMPLATE',
+          "'let-' binding can only be used in 'template' element");
+
   // TODO: Max: Split this error into more smaller, detailed messages.
```
liuming0, line 170: TODO(max): ...

liuming0 (approved):

lib/src/exception_handler/exceptions.dart
mk13: This won't be removed until further changes are made in preserving offsets during micro-expression parsing; this PR won't be handling this.

MichaelRFairhurst (approved):
LGTM!

PR #56, opened 2017-08-25
Publish 0.4.0

MichaelRFairhurst:


pubspec.yaml
```diff
@@ -2,13 +2,13 @@ name: angular_ast
...
 description: Parser and utilities for AngularDart templates
-version: 0.3.1-dev
+version: 0.4.0
 
 environment:
   sdk: '>=1.21.0 <2.0.0'
 
 dependencies:
-  analyzer: '0.30.0+2'
+  analyzer: '0.31.0-alpha.1'
```
bwilkerson, line 11: Not sure who all depends on this package, so I don't know whether this will cause version resolution issues. If this package can still use 0.30, then this should probably be
```
  analyzer: '>=0.30.0+2 <0.32'
```
which will allow it to match any 0.31 version. When pub sees the plugin's constraint it will still match this range.

bwilkerson: Of course, I don't know when `Parser.withoutFasta` was added, so the lower bound might need to be higher because of that.

MichaelRFairhurst: Currently we're the only users of the package.

But this does seem to be causing issues with older SDKs which I think matan wanted it to work with...

I don't see a way around it due to `Parser.withoutFasta`...

MichaelRFairhurst: I'll make this a beta. That should make the less-than-desirable aspects of this versioning, a little bit less...not desirable?

bwilkerson (approved):

bwilkerson (approved):

PR #57, opened 2017-08-30
Accept any amount of event suffixes, ie keyup.ctrl.shift.enter etc

MichaelRFairhurst:
Decided to call the "reductions", for no real good reason. Basically,
they filter down the events, but "filter" sounded to me like something
to linear -- ie, keyup.ctrl.a should be a filter of keyup.a, but that's
not the case. For whatever reason, I felt like reductions didn't imply
this while still being a synonym of filter.

We *could* enforce a, say, 6 limit based on the current modifiers etc
that are allowed for keyup, but that seems arbitrary.

Alternatively, we could try to parse keyup/down modifiers and keys
directly in the ast, rather than leaving it as an array of strings.
That's probably the best long term option.

At the very least, this will allow us to, in the plugin, support
keyup.foo.bar.baz rather than report it as an error. Later on we can
try to validate it in either place, once we've solved the false errors.

### DIFF
--- a/lib/src/ast/event.dart
+++ b/lib/src/ast/event.dart
@@ -69,7 +69,7 @@ abstract class EventAst implements TemplateAst {
   /// Unquoted value being bound to event.
   String get value;
 
-  /// An optional set of postfixes used to filter events that support it.
+  /// An optional list of postfixes used to filter events that support it.
   ///
   /// For example `(keydown.ctrl.space)`.
   List<String> get reductions;
@@ -77,7 +77,7 @@ abstract class EventAst implements TemplateAst {
   @override
   String toString() {
     if (reductions.isNotEmpty) {
-      return '$EventAst {$name.$reductions="$value", Expression=$expression}';
+      return '$EventAst {$name.${reductions.join(',')}="$value", Expression=$expression}';
     }
     return '$EventAst {$name=$value, Expression=$expression}';
   }
@@ -164,11 +164,11 @@ class ParsedEventAst extends TemplateAst
   @override
   int get suffixOffset => suffixToken.offset;
 
-  /// Name `filters` in `(eventName.ctrl.shift.a)`; may be empty.
+  /// Name `reductions` in `(eventName.ctrl.shift.a)`; may be empty.
   @override
   List<String> get reductions {
     final split = _nameWithoutParentheses.split('.');
-    return split.getRange(1, split.length).toList();
+    return split.sublist(1);
   }
 }
 
--- a/lib/src/ast/event.dart
+++ b/lib/src/ast/event.dart
@@ -5,10 +5,13 @@
 import 'package:angular_ast/src/ast.dart';
 import 'package:angular_ast/src/token/tokens.dart';
 import 'package:angular_ast/src/visitor.dart';
+import 'package:collection/collection.dart';
 import 'package:source_span/source_span.dart';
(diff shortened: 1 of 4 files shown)


lib/src/ast/event.dart
```diff
@@ -66,15 +69,15 @@ abstract class EventAst implements TemplateAst {
   /// Unquoted value being bound to event.
   String get value;
 
-  /// An optional postfix used to filter events that support it.
+  /// An optional set of postfixes used to filter events that support it.
```
mk13: Nit: set -> list since order will likely matter.

lib/src/ast/event.dart
```diff
@@ -66,15 +69,15 @@ abstract class EventAst implements TemplateAst {
...
-  String get postfix;
+  /// For example `(keydown.ctrl.space)`.
+  List<String> get reductions;
 
   @override
   String toString() {
-    if (postfix != null) {
-      return '$EventAst {$name.$postfix="$value", Expression=$expression}';
+    if (reductions.isNotEmpty) {
+      return '$EventAst {$name.$reductions="$value", Expression=$expression}';
```
mk13: Perhaps ${reductions.join('.')} since $reductions will give something like "['keydown','space']".

EDIT: Actually I'll leave this decision up to you; having explicit '[' and ']' might be nice for toString() since HumanizingVisitor already shows the proper representation in literal form.

lib/src/ast/event.dart
```diff
@@ -153,20 +156,19 @@ class ParsedEventAst extends TemplateAst
...
   @override
   int get prefixOffset => prefixToken.offset;
 
-  /// Offset of `)` suffix in `(eventName.postfix)`.
+  /// Offset of `)` suffix in `(eventName.reductions)`.
   @override
   int get suffixOffset => suffixToken.offset;
 
-  /// Name `postfix` in `(eventName.postfix)`; may be `null` to have no value.
+  /// Name `filters` in `(eventName.ctrl.shift.a)`; may be empty.
```
mk13: I'm not sure where 'filters' came from? Perhaps from an older version of your code when you called it filters?

lib/src/ast/event.dart
```diff
@@ -153,20 +156,19 @@ class ParsedEventAst extends TemplateAst
...
   @override
-  String get postfix {
+  List<String> get reductions {
     final split = _nameWithoutParentheses.split('.');
-    assert(split.length < 2);
-    return split.length > 1 ? split[1] : null;
+    return split.getRange(1, split.length).toList();
```
mk13: split.subList(1) is probably a more succinct way.

mk13 (approved):
Sorry for the delay! LGTM.

lib/src/ast/event.dart
MichaelRFairhurst: Done!

lib/src/ast/event.dart
MichaelRFairhurst: Agreed! Done

lib/src/ast/event.dart
MichaelRFairhurst: Yep, exactly. Fixed!

lib/src/ast/event.dart
MichaelRFairhurst: Ooh, very nice. Done!

